PR #100, opened 2020-05-28
Rename parameters and minor edits to plot.bambu function and updated readme and vignettes, added the output to gtf and txt functions, updated PrepareAnnotationFromGTF function

cying111:
I have 
- updated the argument names to more user friendly names
- updated corresponding readme and vignettes files 
- updated plot.bambu functions for better annotation
- implemented messages to show processing status when verbose is set to FALSE, with progressbar for bplapply 
- updated the condition test to a more generalised version for CPU distribution for parallel processing

- modified the outputFunctions
- updated PrepareAnnotationFromGTF function


R/readWrite.R
```diff
@@ -0,0 +1,97 @@
...
+#' \itemize{
+#'   \item TXNAME specifying prefix for new gene Ids (genePrefix.number), defaults to empty
+#'   \item GENEID indicating whether filter to remove read classes which are a subset of known transcripts(), defaults to TRUE
+#'   }
+#' @export
+readFromGTF <- function(file){
+  if (missing(file)){
+    stop('A GTF file is required.')
+  }else{
+    data=read.delim(file,header=FALSE)
```
jonathangoeke, line 83:
```suggestion
    data=read.delim(file,header=FALSE, comment.char='#')
```
(do we need to remove # lines for gtf?)

PR #116, opened 2020-06-16
strip off "chr" on seqnames if exist

yuukiiwa:
I have compared between the working gtf file and the not working one, and it seems like the one that's not working has "chr" before the chromosome number. I have run using this updated function on RStudio without errors:
Screen Shot 2020-06-16 at 5 07 24 PM


yuukiiwa:
I downloaded the Homo_sapiens_NCBI_GRCh38 igenome directory to my computer, and using the genome.fa and genes.gtf files ran successfully:
samplesheet.csv
Screen Shot 2020-06-16 at 7 04 18 PM
run
Screen Shot 2020-06-16 at 7 02 24 PM

PR #138, opened 2020-07-09
update RData names from NanoporeRNASeq

yuukiiwa:
The RData names from NanoporeRNASeq were changed, so I change it here too. The Google Doc is updated.


yuukiiwa:
The Google Doc is updated. Here are the changes made to the vignette:
- added title, authors, table of contents with internal links, and data preprocessing descriptions.
- included install_github for NanoporeRNASeq
- added path to fasta file and the BSgenome library as two options for loading genome sequences
- changed seExtended to se and se to seUnextended
- replaced colData from se with sample_info$cellLine and $sample_id columns
- add in the apeglm library for running running lfcShrink using apeglm

PR #150, opened 2020-07-29
minor fixes on documentation

yuukiiwa:
@cying111, here are some minor fixes on documentation. I also changed install() to BiocManager::install() as install() may not be able to find Bioconductor packages.


R/annotationFunctions.R
```diff
@@ -56,7 +56,7 @@ prepareAnnotations <- function(x) {
 #'   \item GENEID indicating whether filter to remove read classes which are a subset of known transcripts(), defaults to TRUE
 #'   \item eqClass specifying minimun read count to consider a read class valid in a sample, defaults to 2
 #'   }
-#' @noRd
+#' @export
```
cying111, line 59: i think the function has been changed such that this function will no longer be exported

yuukiiwa: I have reverted it back to @noRd, and now prepareAnnotationFromGTF() is not exported from NAMESPACE.

PR #177, opened 2020-12-21
update annotateSpliceOverlapByDist

yuukiiwa:
Hi @jonathangoeke, I have reforked bambu and updated the files I changed and added for annotateSpliceOverlapByDist().
Thanks!


R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
-# Note: several of the functions in this file are adopted from 
-# the GenomicAlignments package (Author: Hervé Pagès, Valerie Obenchain,
-# Martin Morgan)
-# License Artistic-2.0
+# Note: several of the functions in this file are adopted from the GenomicAlignments package (Author: Hervé Pagès, Valerie Obenchain, Martin Morgan)
+# License	Artistic-2.0
 # https://doi.org/doi:10.18129/B9.bioc.GenomicAlignments
 
-## examples for test purposes
```
jonathangoeke, line 7: Can you document the test data somewhere using these descriptions? Where would be the best place to document the data?

R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
-# Note: several of the functions in this file are adopted from 
-# the GenomicAlignments package (Author: Hervé Pagès, Valerie Obenchain,
-# Martin Morgan)
-# License Artistic-2.0
+# Note: several of the functions in this file are adopted from the GenomicAlignments package (Author: Hervé Pagès, Valerie Obenchain, Martin Morgan)
```
jonathangoeke, line 1: why this change? Wouldn't it be better to use the original line wrapped version for Bioconductor compatibility?

R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
...
-  subjectStartExonStartExtension <- start(subjectStart.querySplice.Rng)<end(qSpRngList) & end(subjectStart.querySplice.Rng)>end(qSpRngList) & start(qSpRngList)< start(subjectStart.querySplice.Rng)
-  subjectStartExonStartExtension <- tapply(subjectStartExonStartExtension, subjectStartEnd.querySplice.IdMap, sum)
+  olap
+}
 
-  subjectStartExonEndExtension <- end(subjectStart.querySplice.Rng)>start(qSpRngList) & start(subjectStart.querySplice.Rng)<start(qSpRngList) & end(qSpRngList)> end(subjectStart.querySplice.Rng)
-  subjectStartExonEndExtension <- tapply(subjectStartExonEndExtension, subjectStartEnd.querySplice.IdMap, sum)
+#' annotate splice overlap by distance
+#' @noRd
+annotateSpliceOverlapsByDist <-function(query, subject) {
```
jonathangoeke, line 101: choose a more meaningful name? what about compareTranscripts()?
Can you add the documentation for this function, including the example with the example data?

jonathangoeke: it might be nice to illustrate how to use this in the bambu workflow. Do you want to create an issue for that?

R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
...
+                                                                     queryFullRng, annotatedTable$alternativeLastExon)
+  #intron retention subject and query
+  annotatedTable$intronRetention.subject <- annotateIntronRetent(querySpliceRng, subjectFullRng)
+  annotatedTable$intronRetention.query <- annotateIntronRetent(subjectSpliceRng ,queryFullRng)
+  #exon skipping query and subject
+  annotatedTable$exonSkipping.query <- annotateExonSkip(querySpliceRng, subjectFullRng,
+                                                        subjectStartRng, subjectEndRng)
+  annotatedTable$exonSkipping.subject <- annotateExonSkip(subjectSpliceRng ,queryFullRng,
+                                                          queryStartRng, queryEndRng) 
+  # exon 3' splice site
```
jonathangoeke, line 144: what about excon 5' splice site? Will both be done in this function? Then that might be good to include in the comment

R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
...
+                                        annotatedTable$strand)
+  annotatedTable <- cbind(annotatedTable, exonSpliceTable)
+  return(annotatedTable)
+}
 
-  subjectEndExonStartExtension <- start(subjectEnd.querySplice.Rng)<end(qSpRngList) & end(subjectEnd.querySplice.Rng)>end(qSpRngList) & start(qSpRngList)< start(subjectEnd.querySplice.Rng)
-  subjectEndExonStartExtension <- tapply(subjectEndExonStartExtension, subjectStartEnd.querySplice.IdMap, sum)
+#' extract strand from GRangesList
+#' @noRd
+getStrandFromGrList <- function(grl) {
```
jonathangoeke, line 154: can you document this function, including an example?

R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
...
-  subjectEndExonStartExtension <- start(subjectEnd.querySplice.Rng)<end(qSpRngList) & end(subjectEnd.querySplice.Rng)>end(qSpRngList) & start(qSpRngList)< start(subjectEnd.querySplice.Rng)
-  subjectEndExonStartExtension <- tapply(subjectEndExonStartExtension, subjectStartEnd.querySplice.IdMap, sum)
+#' extract strand from GRangesList
+#' @noRd
+getStrandFromGrList <- function(grl) { 
+  return(unlist(strand(grl), use.names = F)[cumsum(elementNROWS(grl))]) 
+  }
+
+#' start/end ranges pre-processing                                 
+expandRanges <- function(unexpandedRng,targetRng){
```
jonathangoeke, line 159: Here we might not need to add the full documentation, but can you add comments on what this function does?

R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
...
+  processedRng <- rep(unexpandedRng,elementNROWS(targetRng))
+  mcols(processedRng)$IdMap <- rep(1:length(unexpandedRng),elementNROWS(targetRng))
+  mcols(processedRng)$matchRng <- unlist(targetRng, use.names=F)
+  return (processedRng)
+}
 
-  subjectEndExonEndExtension <- end(subjectEnd.querySplice.Rng)>start(qSpRngList) & start(subjectEnd.querySplice.Rng)<start(qSpRngList) & end(qSpRngList)> end(subjectEnd.querySplice.Rng)
-  subjectEndExonEndExtension <- tapply(subjectEndExonEndExtension, subjectStartEnd.querySplice.IdMap, sum)
+#' splice ranges pre-processing
+expandRangesList <- function(unexpandedRng,targetRng){
```
jonathangoeke, line 167: also add some comments on what this function does

R/utility_spliceHelper_functions.R
```diff
@@ -1,598 +1,572 @@
...
+  mcols(processedRng)$matchRng <- unlist(rep(targetRng, times= elementNROWS(unexpandedRng)), use.names=F)
+  return (processedRng)
+}
 
-  startEndTable$exon5Prime <- exon5Prime
-  startEndTable$exon5Prime[startEndTable$strand!='-'] <- startEndTable$exon5Prime[startEndTable$strand!='-'] - subjectStartExonStartExtension[startEndTable$strand!='-']
-  startEndTable$exon5Prime[startEndTable$strand=='-'] <- exon3Prime[startEndTable$strand=='-'] - subjectStartExonEndExtension[startEndTable$strand=='-']
+#' alternative start/end exon
+#' @noRd
+alternativeStartEndExon <- function(queryRng, subjectRng){
```
jonathangoeke, line 176: also add a comment here

jonathangoeke (changes requested):
Thanks Yuk Kei! Please document the functions, or add comments to describe what the functions are doing. Can you also check that no functions/lines are changed that are not directly relevant to the new code?

R/utility_spliceHelper_functions.R
yuukiiwa: I have added the document of the test data back on top of the compareTranscripts() function.

R/utility_spliceHelper_functions.R
yuukiiwa: I have reverted this back to the GoekeLab/bambu-bioc-dev version.

R/utility_spliceHelper_functions.R
yuukiiwa: I have documented this function with a description and an example.

R/utility_spliceHelper_functions.R
yuukiiwa: I have added a description to this function.

R/utility_spliceHelper_functions.R
yuukiiwa: I have added a description to this function.

R/utility_spliceHelper_functions.R
yuukiiwa: I have added a description to this function.

R/utility_spliceHelper_functions.R
yuukiiwa: I have added an [issue](https://github.com/GoekeLab/bambu/issues/178) to address this.

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,236 @@
...
+# query <- readRDS(system.file("extdata", 
+#     "annotateSpliceOverlapByDist_testQuery.rds",
+#     package = "bambu"))
+# subject <- readRDS(system.file("extdata", 
+#        "annotateSpliceOverlapByDist_testSubject.rds",
+#        package = "bambu"))
+############################################################
+#' annotate splice overlap by distance
+#' @noRd
+compareTranscripts <-function(query, subject) {
```
jonathangoeke, line 33: Can you add a complete documentation for the user for this function? This should be accessbile by users, I think it will be very useful!

R/utility_spliceHelper_functions.R
jonathangoeke: thanks! Can you add Usage, Arguments, Value? See ?mean for a short documentation example

R/utility_spliceHelper_functions.R
jonathangoeke: can you add a description to all the new functions? Please include what the function does

R/utility_spliceHelper_functions.R
```diff
@@ -572,6 +588,43 @@ dropGrangesListElementsByWidth <- function(grangesList, minWidth = 5,
...
+#' @param range IRangesList with elements required to be ordered by coordinates
+#' @param stand strand
+#' @noRd
+selectEndExonFromRangesList <- function(range, strand){
+  endExonsSet <- as.numeric(cumsum(elementNROWS(range)))
+  firstExons <- c(1, endExonsSet[-(length(endExonsSet))]+1)
+  endExonsSet[strand == "-"] <- firstExons[strand == "-"]
+  return(unlist(range, use.names = FALSE)[endExonsSet])
+}
+####the ones with relist()
```
jonathangoeke, line 612: why are these included here, do we plan to use them later?

jonathangoeke (changes requested):
Thanks! Can you include documentation for the new functions?

R/utility_spliceHelper_functions.R
yuukiiwa: I have updated the comment to include 5' splice sites

R/utility_spliceHelper_functions.R
yuukiiwa: Descriptions that include what the function does are added to all new functions.

R/utility_spliceHelper_functions.R
yuukiiwa: I have removed them

R/utility_spliceHelper_functions.R
yuukiiwa: I added a full documentation with arguments, values, and examples.

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,334 @@
...
+  exonSpliceTable <- annotateExonSplice(querySpliceRng, subjectFullRng,
+                                        subjectStartRng, subjectEndRng,
+                                        annotatedTable$strand)
+  annotatedTable <- cbind(annotatedTable, exonSpliceTable)
+  return(annotatedTable)
 }
 
+#' extract strand from GRangesList
+#' @description This function takes a GRangesList and
+#' get strand information
```
jonathangoeke, line 119:
```suggestion
#' returns a vector with the strand for each list entry.
#' This function assumes that all elements for each list
#' entry have the same strand.
```

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,334 @@
...
+  processedRng <- rep(unlist(rglist, use.names=F),rep(elementNROWS(target),times=elementNROWS(rglist)))
+  mcols(processedRng)$IdMap <- rep(1:length(rglist),elementNROWS(rglist)*elementNROWS(target))
+  mcols(processedRng)$matchRng <- unlist(rep(target, times= elementNROWS(rglist)), use.names=F)
+  return (processedRng)
+}
+
+#' alternative start/end exon
+#' @description This function checks whether an 
+#' alternative start/end exon is used by overlapping
+#' the exon ranges of matching transcripts.
```
jonathangoeke, line 193:
```suggestion
#' the exon ranges of the first (or last) exons of matching transcripts.
```

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,334 @@
...
+#' @noRd
+alternativeStartEndExon <- function(queryRng, subjectRng){
+  return(!poverlaps(queryRng, subjectRng))
+}
+
+#' alternative TSS distance
+#' @description This function calculates the distance
+#' of an alternative TSS by comparing the start
+#' coordinates of the start exon ranges of matching
+#' transcripts if an alternative first exon is not used.
```
jonathangoeke, line 203:
```suggestion
#' transcripts. If an alternative first exon is used the distance is set to 0.
```

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,334 @@
...
+  alternativeTSS <- alternativeTSS * !alternativeFirstExon
+  return (alternativeTSS)
+}
+
+
+#' alternative TES distance
+#' @description This function calculates the distance
+#' of an alternative TES by comparing the end
+#' coordinates of the end exon ranges of matching
+#' transcripts if an alternative last exon is not used.
```
jonathangoeke, line 219:
```suggestion
#' transcripts. If an alternative last exon is used the distance is set to 0.
```

jonathangoeke (changes requested):
just some minor edits!

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,336 @@
...
+#' @params grl a GRangesList
+#' @return an Rle object with strand information
+#' @examples
+#' query <- readRDS(system.file("extdata", 
+#'     "annotateSpliceOverlapByDist_testQuery.rds",
+#'     package = "bambu"))
+#' strand <- as.character(getStrandFromGrList(query))
+#' @noRd
+getStrandFromGrList <- function(grl) { 
+  return(unlist(strand(grl), use.names = F)[cumsum(elementNROWS(grl))])
```
cying111, line 132: hi @yuukiiwa, you can use this line to get the strand 
`unique(unlist(strand(grl))`

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,336 @@
...
+  return(!poverlaps(queryRng, subjectRng))
+}
+
+#' alternative TSS distance
+#' @description This function calculates the distance
+#' of an alternative TSS by comparing the start
+#' coordinates of the start exon ranges of matching
+#' transcripts. If an alternative first exon is used the distance is set to 0.
+#' @noRd
+calculateTSSdistance <- function(queryStartRng, subjectStartRng,
```
cying111, line 207: Hi @yuukiiwa, I am not sure if this is a good or bad idea, I feel that calculateTSSdistance and calculateTESdistance can potentially be collapsed into one function, as they are highly similar except the internal function used is slightly different(one is using start, the other is using end). If I were you, I probably will create a function call `calculateTerminalDistance <- function(queryTerminalExonRng, ...alternativeTerminalExonm, strand, direction = "Start")`, and use the direction to determine which function to be used

R/utility_spliceHelper_functions.R
```diff
@@ -572,6 +688,28 @@ dropGrangesListElementsByWidth <- function(grangesList, minWidth = 5,
     return(relist(unlistedExons, partitioning))
 }
 
+#' Function that selects the first exon from an IRangesList object
+#' @param range IRangesList with elements required to be ordered by coordinates
+#' @param stand strand
+#' @noRd
+selectStartExonFromRangesList <- function(range, strand){
```
cying111, line 695: similarly for these two functions: `selectStartExonFromGrangesList` and `selectEndExonFromGrangesList`, they look very similar to me, they can potentially be one function and they don't even need to have an extra variable I feel.

R/utility_spliceHelper_functions.R
```diff
@@ -20,220 +20,336 @@
...
-
-  startEndTable$exon5Prime <- exon5Prime
-  startEndTable$exon5Prime[startEndTable$strand!='-'] <- startEndTable$exon5Prime[startEndTable$strand!='-'] - subjectStartExonStartExtension[startEndTable$strand!='-']
-  startEndTable$exon5Prime[startEndTable$strand=='-'] <- exon3Prime[startEndTable$strand=='-'] - subjectStartExonEndExtension[startEndTable$strand=='-']
-
-  startEndTable$exon3Prime <- exon3Prime
-  startEndTable$exon3Prime[startEndTable$strand!='-'] <- startEndTable$exon3Prime[startEndTable$strand!='-'] - subjectEndExonEndExtension[startEndTable$strand!='-']
-  startEndTable$exon3Prime[startEndTable$strand=='-'] <- exon5Prime[startEndTable$strand=='-'] - subjectEndExonStartExtension[startEndTable$strand=='-']
-  return(startEndTable)
+  strand <- as.character(getStrandFromGrList(query))
```
cying111, line 69: Hi @yuukiiwa , given that the strand can be obtained with a single line, it actually brought up another question that I have, is it really necessary to have a function call `getStrandFromGrgList`?

cying111 (commented):
Hi @yuukiiwa, this looks good to me! I had a few comments regarding the sub-functions, I generally feel that some of the sub-functions are quite similar, so instead of having one function for each case, have just one function with additional variable to indicate purpose might be a better way of forward, For details, please see my comments! Thanks!

R/utility_spliceHelper_functions.R
jonathangoeke: > Hi @yuukiiwa, this looks good to me! I had a few comments regarding the sub-functions, I generally feel that some of the sub-functions are quite similar, so instead of having one function for each case, have just one function with additional variable to indicate purpose might be a better way of forward, For details, please see my comments! Thanks!

Thanks for the review @cying111! The getStrandFromGrList was an idea from me, as this can be helpful for some other parts in the code where I used to store the strand so that I don't have to recalculate it. This calculation is fast, which makes it more efficient than storing the strand, so this might help make the code much simpler

yuukiiwa: Hi @cying111, I do agree that the getStrandFromGrList function can be expressed as a command within the main compareTranscripts() function. The reason why we made this function was to make the main compareTranscripts() function cleaner to make it easier to be interpreted. If you think it is unnecessary I can replace the current function with the command.

R/utility_spliceHelper_functions.R
yuukiiwa: Hi @cying111, I do agree that the getStrandFromGrList function can be expressed as a command within the main compareTranscripts() function. The reason why we made this function was to make the main compareTranscripts() function cleaner to make it easier to be interpreted. If you think it is unnecessary I can replace the current function with the command.

R/utility_spliceHelper_functions.R
yuukiiwa: With this direction variable, we then need an if...else... to separate the operations for TSS and TES, which will not be as time efficient.

R/utility_spliceHelper_functions.R
cying111: I think from the time efficient view as @jonathangoeke mentioned, then you can leave it as it is then!

R/utility_spliceHelper_functions.R
cying111: You are right maybe! If so, I guess you can leave this function as it is as well!

R/utility_spliceHelper_functions.R
yuukiiwa: Hi @cying111, yes they are very similar, and in fact, the first, second, and last line of the functions are the same. I don't think I can merge them into one function because the direction of the third lines of the two functions are different, where for selectStartExonFromRangesList(), the variable from the second line of the function is used while selectEndExonFromRangesList() returns the variable from the first line of the function.

yuukiiwa: I guess I will also keep this one.

R/utility_spliceHelper_functions.R
cying111: Sorry for coming back to this again, I realized that there is some advanced way that you can do to avoid the `if` if you feel it's less time efficient?:
direction_names <- c("Start","End")
if direction = "Start", the first line can be like this 
`.... <- get(tolower(direction))()` 
the second line will be 
"....[strand == "-"]  <- get(tolower(setdiff(direction_names, direction)))()" 

'get' is a function to use for variable function names

R/utility_spliceHelper_functions.R
cying111: I see, for this one, I would probably still combine as they are all numeric vectors, so the comparison with if would not slow down the function much, I feel. It would still be good to have less functions that with similar internal designs though.

yuukiiwa: I have combined the two functions.

R/utility_spliceHelper_functions.R
yuukiiwa: I have merged the two functions.

jonathangoeke (approved):

PR #185, opened 2021-01-12
update to file naming and structure: bambu-processReads and utility functions

jonathangoeke:
- created utility function files for bambu.processReads according to the major tasks (junction table creation, error correction, read class construction).
- seqlevel consistency checks are now performed based on the reference genome sequence that is provided. seqlevels and ranges which are not present in the genome sequence are dropped.
- bugfix: strand correction for reads now corrects for reads (previously was likely to be disabled)


R/bambu-processReads.R
```diff
@@ -53,20 +53,49 @@ bambu.processReads <- function(reads, readClass.file, annotations, genomeSequenc
 #' @noRd
 bambu.processReadsByFile <- function(bam.file, genomeSequence, annotations,
```
cying111, line 54: This function seems to have a length of over 50 lines? Maybe can slightly reduce it a bit?

R/bambu-processReads_utilityCreateJunctionTables.R
```diff
@@ -0,0 +1,260 @@
...
+    mcols(gr)$id <- rep(mcols(x)$id, elementNROWS(gaps))
+  return(gr)
+}
+
+
+
+#' Create Junction tables from unlisted junction granges
+#' @importFrom BiocParallel bppram bpvec
+#' @noRd
+createJunctionTable <- function(unlisted_junctions,
```
cying111, line 71: This function also has a length slightly over 50 lines?

cying111 (changes requested):
Hi @jonathangoeke, thanks for the updates, the changes look fine to me, only two functions have slightly length over 50 lines, maybe can consider to shorten the two functions a bit?

R/bambu-processReads.R
jonathangoeke: the function is now less than 50 lines

R/bambu-processReads_utilityCreateJunctionTables.R
jonathangoeke: I removed a check for constent seqlevels which was redundant, the function is now below 50 lines

jonathangoeke:
Thanks @cying111 for looking through this. I changed the 2 functions to reduce the line number

cying111 (approved):

Drakyrie (approved):

PR #189, opened 2021-01-14
comparison of quant and isore branches

jonathangoeke:


jonathangoeke:
created pull request for comparison

PR #195, opened 2021-01-22
created assignGeneIds function

jonathangoeke:
I added a function to assign gene ids based on transcript overlaps. We had a function like this which was used in extendAnnotations, but that function is slow and memory intensive. The new function is now in the processReads module so it can be applied for each sample read class. Unlike the function in extendAnnotations, the new function does not assign read classes to genes if the strand is different, * is not considered unstranded.

This pull request needs some edits in terms of formatting and testing of this function, so @Drakyrie you can also merge this into your own branch first, make some edits, and merge it into the dev branch later.


andredsim:
This was included and modified in #200

PR #204, opened 2021-04-26
Rename files for extendAnnotations functions

jonathangoeke:
- renamed files to be consistent with other file names for the extendAnnotations code
- reordered functions so that they appear in the way they are called from within the code


andredsim (approved):
Runs fine on my test cases. Approved from my end.

PR #207, opened 2021-04-30
Changes to improve code stability

keithyjy:
Some fixes to make the code base more stable. The first change is to remove the use of get() which looks into environment variables (prone to crashes if a variable of the same name exists) and directly references the functions that should be used instead. Changes are in bambu-processReads_utilityJunctionErrorCorrection.R


andredsim (approved):
Not a functional change, but stops the code from crashing if we ever decide to use start or end as variable names earlier than this

cying111 (approved):

PR #218, opened 2021-08-11
noAnno_patch

andredsim:
Improve the performance of Bambu when large amounts of annotations are missing.

Gene Model now only has 5 rounds of xgboost to stop overfitting on the smaller amount of data
TRUE genes are now if any of their gene members are compatible with at least one annotation
Gene read prop is calculated under the bambu gene definition and does not use annotations
Annotations are now filtered by a geneScore of 0.2
assignNewGeneIds() has been removed and all calls of it have been replaced by the faster assignGeneIds()

### DIFF
--- a/R/bambu-extendAnnotations-utilityExtend.R
+++ b/R/bambu-extendAnnotations-utilityExtend.R
@@ -3,10 +3,11 @@
 #' @noRd
 isore.extendAnnotations <- function(combinedTranscripts, annotationGrangesList,
                                     remove.subsetTx = TRUE,
-                                    min.sampleNumber = 1, min.exonDistance = 35, min.exonOverlap = 10,
+                                    min.sampleNumber = 1, max.txFDR = 0.1, min.exonDistance = 35, min.exonOverlap = 10,
                                     min.primarySecondaryDist = 5, min.primarySecondaryDistStartEnd = 5, 
                                     prefix = "", verbose = FALSE){
   filterSet <- filterTranscriptsByRead(combinedTranscripts, min.sampleNumber)
+  filterSet <- filterTranscriptsByFDR(combinedTranscripts, filterSet, max.txFDR)
   if (any(filterSet, na.rm = TRUE)) {
     transcriptsFiltered <- combinedTranscripts[filterSet,]
     group_var <- c("intronStarts","intronEnds","chr","strand","start","end",
@@ -53,7 +54,28 @@ filterTranscriptsByRead <- function(combinedTranscripts, min.sampleNumber){
       combinedTranscripts$NSampleReadProp >= min.sampleNumber) & (
         combinedTranscripts$NSampleGeneScore >= min.sampleNumber) & (
           combinedTranscripts$NSampleTxScore >= min.sampleNumber)
-  # filter based on read count and transcript usage
+  return(filterSet)
+}
+
+#' filter transcripts by FDR
+#' @noRd
+filterTranscriptsByFDR <- function(combinedTranscripts, filterSet, max.txFDR){
+  if (any(filterSet, na.rm = TRUE)) {
+    combinedTranscripts <- combinedTranscripts[filterSet,]
+  } else return(filterset)
+
+  # calculate and filter by FDR
+  combinedTranscripts$equal[is.na(combinedTranscripts$equal)] = FALSE
+  if(sum(combinedTranscripts$equal, na.rm = TRUE)<50 | 
+      sum(!combinedTranscripts$equal, na.rm = TRUE)<50){
+        txFDR = 1 - combinedTranscripts$maxTxScore
+        warning("Less than 50 TRUE or FALSE read classes for precision stabilization. 
+        Filtering by prediction score instead")
+  } else txFDR = calculateFDR(combinedTranscripts$maxTxScore, combinedTranscripts$equal)
+  index = which(filterSet)
+  filterSet[index] = filterSet[index] & (txFDR <= max.txFDR)
+  # remove equals to prevent duplicates when merging with anno
+  filterSet[index] = filterSet[index] & (!combinedTranscripts$equal)
   return(filterSet)
 }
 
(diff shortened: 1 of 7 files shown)


R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -44,17 +44,16 @@ isore.extendAnnotations <- function(combinedTranscripts, annotationGrangesList,
...
     #filter by read counts
     filterSet <- combinedTranscripts$NSampleReadCount >= min.sampleNumber & (
-          combinedTranscripts$NSampleReadProp >= min.sampleNumber) & (
-            combinedTranscripts$NSampleGeneScore >= min.sampleNumber) & (
-              combinedTranscripts$NSampleTxScore >= min.sampleNumber)
-    combinedTranscripts = combinedTranscripts[filterSet,]
+        combinedTranscripts$NSampleTxScore >= min.sampleNumber) & (
+        combinedTranscripts$NSampleReadProp >= min.sampleNumber)
   }  
+  combinedTranscripts = combinedTranscripts[filterSet,]
```
jonathangoeke, line 56: that line might break if `(nrow(combinedTranscripts) <= 0)` as filterSet is not defined outside the if?

R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -548,74 +547,16 @@ assignGeneIDexonMatch <- function(rowDataCombined, exonRangesCombined,
...
-
-#' Assign New Gene with Gene Ids
-#' @param exByTx exByTx
-#' @param prefix prefix, defaults to empty
-#' @param minoverlap defaults to 5
-#' @param ignore.strand defaults to FALSE
-#' @importFrom dplyr as_tibble %>% filter select ungroup inner_join mutate
-#'     distinct if_else arrange tibble
-#' @noRd
-assignNewGeneIds <- function(exByTx, prefix = "", minoverlap = 5,
```
jonathangoeke, line 572: I think this might be the only function that uses "prefix"? we can remove prefix from the argument list then?

R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -331,31 +353,34 @@ calculateDistToAnnotation <- function(exByTx, exByTxRef, maxDist = 35,
...
+                            primarySecondaryDist, exByTx = exByTx,
+                            setTMP = setTMPRest, DistCalculated = TRUE)
+      }
     }
-  }
-  txToAnTableFiltered <- rbind( txToAnTableFiltered,
-                                txToAnTableRest, txToAnTableRestStartEnd ) %>% ungroup()
+    txToAnTableFiltered <- rbind( txToAnTableFiltered,
+                                  txToAnTableRest, txToAnTableRestStartEnd ) %>% ungroup()
+  } else txToAnTableFiltered %>% ungroup()
```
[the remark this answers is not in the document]
andredsim: I can't find this line, might be outdated?

R/bambu-extendAnnotations-utilityExtend.R
andredsim: dealt with in f60dd70a3c2427ec8bcc64e0bf048d047b31c496

R/bambu-extendAnnotations-utilityExtend.R
andredsim: bfec4850170b708dcf28bbf18743ec8b99da970c
At first i was going to remove prefixes but I realized there is a potential use case where a user might rerun bambu with bambu annotations which would mean the gene and tx ids would be duplicated and cause issues in the gtf file. By providing a unique prefix when running bambu it stops that duplication from occuring.
Therefore I returned the prefix to assignGeneIdsByReference(). This should have no functional change for regular use.

R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -44,17 +44,17 @@ isore.extendAnnotations <- function(combinedTranscripts, annotationGrangesList,
...
   if (nrow(combinedTranscripts) > 0){
     #filter by read counts
     filterSet <- combinedTranscripts$NSampleReadCount >= min.sampleNumber & (
-          combinedTranscripts$NSampleReadProp >= min.sampleNumber) & (
-            combinedTranscripts$NSampleGeneScore >= min.sampleNumber) & (
-              combinedTranscripts$NSampleTxScore >= min.sampleNumber)
-    combinedTranscripts = combinedTranscripts[filterSet,]
+        combinedTranscripts$NSampleTxScore >= min.sampleNumber) & (
+        combinedTranscripts$NSampleReadProp >= min.sampleNumber)
   }
```
cying111, line 56:
```suggestion
 combinedTranscripts <- combinedTranscripts[filterSet,]
   }  
```

R/bambu-extendAnnotations-utilityExtend.R
cying111:
```suggestion

```

R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -44,17 +44,17 @@ isore.extendAnnotations <- function(combinedTranscripts, annotationGrangesList,
...
 }
 
-#' filter transcripts by read counts and NDR
+#' filter transcripts by read counts
 #' @noRd
-filterTranscripts <- function(combinedTranscripts, min.sampleNumber, max.txNDR){
+filterTranscripts <- function(combinedTranscripts, min.sampleNumber){
+  filterSet = NULL
```
cying111, line 50: when filterSet is NULL, I think sometimes, `combinedTranscripts[filterSet,]` will throw error as there is nothing to select. Suggest to remove 
```suggestion
```

cying111 (approved):

cying111:
Hi @andredsim , the new changes look good to me and testing with random samples works fine, quantification results remain very similar with before this PR.

R/bambu-extendAnnotations-utilityExtend.R
andredsim: If filterSet is NULL then combinedTranscripts will be a table with 0 rows similar to if everything is filtered out. This means the same condition can be used to check if its empty because there were no transcripts or if there were no valid transcripts in isore.extendAnnotations. If combinedTranscripts could sometimes == NULL then 
> if (nrow(NULL)<0) will cause an error.

PR #238, opened 2022-02-04
Small Bug Fixes and Unit Tests

andredsim:


R/bambu.R
```diff
@@ -127,6 +127,10 @@ bambu <- function(reads = NULL, rcFile = NULL, rcOutDir = NULL,
             readClass.outputDir = rcOutDir, genomeSequence = genome)
     if(!is.null(reads)) genomeSequence <- checkInputSequence(genome)
     isoreParameters <- setIsoreParameters(isoreParameters = opt.discovery)
+
+    #below line is to be compatible with earlier version of running bambu
+    if(!is.null(isoreParameters$max.txNDR)) NDR = isoreParameters$max.txNDR
+
```
cying111, line 133: it will be a bit confusing to users I guess if we have two places to set the NDR threshold, I would suggest to just move NDR out of the isoreParameters. It is not exactly a new feature, right? We probably also need to update the documentation for this as well

cying111 (commented):
Hi @andredsim , thanks for fixing the bugs. The changes look good to me, only got one small comment for the NDR threshold setting, should we just move the NDR outside of the isoreParameters and also update documentation(readme, vignettes) to reflect this change so that users can follow? Thank you

PR #290, opened 2022-07-18
remove theta from the final se output list object

cying111:


jonathangoeke (commented):
looks good. If results are identical can be merged

PR #296, opened 2022-07-19
add unit test for writeBambuOutput and update bambu output

lingminhao:
Here I add some unit tests for the writeBambuOutput() function. 
It tests whether: 
- the prefix is as expected 
- the output files go to the correct output path
- the file extensions are correct
- the column names for the count outputs include "GENEID" and/or "TXNAME".
- the sample columns for count output files are numeric.

Additionally,  since the theta column was removed from the Bambu output, some of the internal/external data would also need to be updated.

### DIFF
--- a/tests/testthat/test_readWrite.R
+++ b/tests/testthat/test_readWrite.R
@@ -1,5 +1,31 @@
 context("Generate GTF file from summarizedExperiment object")
 
+# test for the function: writeBambuOutput
+test_that("the output files have correct prefix, file format and all of them are placed correctly in the given output path", {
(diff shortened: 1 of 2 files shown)


tests/testthat/test_readWrite.R
```diff
@@ -1,5 +1,31 @@
 context("Generate GTF file from summarizedExperiment object")
 
+# test for the function: writeBambuOutput
+test_that("the output files have correct prefix, file format and all of them are placed correctly in the given output path", {
+    se <- readRDS(system.file("extdata", "seOutput_SGNex_A549_directRNA_replicate5_run1_chr9_1_1000000.rds", package = "bambu"))
+    gtf.file <- system.file("extdata", "Homo_sapiens.GRCh38.91_chr9_1_1000000.gtf", package = "bambu")
```
jonathangoeke, line 6: why is this needed?

PR #308, opened 2022-08-22
Update xgboost saved models

andredsim:


jonathangoeke (commented):
check that model txt files are correct
reduce sie of default model?

jonathangoeke (approved):

PR #314, opened 2022-09-14
update vignettes

cying111:


vignettes/bambu.Rmd
```diff
@@ -156,46 +154,83 @@ annotation, and then quantify the transcript expression from these extended
...
+Additionally there are other thresholds that advanced users can access through opt.discovery when running *bambu* (see arguments).
 
 ### Visualise results {#visualise-results}               
 *bambu* provides functions to visualise and explore the results. When multiple
 samples are used, we can visualise the correlation and clustering of all
 samples with a heatmap:
 ```{r, fig.width = 8, fig.height = 6}
 library(ggplot2)
-plotBambu(se, type = "heatmap")
+plotBambu(se.multiSample, type = "heatmap")
```
andredsim, line 225: Error in loadNamespace(x) : there is no package called ‘ComplexHeatmap’

vignettes/bambu.Rmd
```diff
@@ -204,25 +239,31 @@ individual genes. Here we look at gene ENSG00000099968 and visualise the
...
+a comparison of the 2 cell lines:
+```{r}
+colData(se.multiSample)$condition <- as.factor(SGNexSamples$cellLine)
+```
+
+To obtain the accurate gene expression estimates which uses all reads that can be assigned to each gene (including reads that are incompatible with all existing annotations) you can run the following command. Looking at the output, we can see there are novel genes identified as well
 ```{r}
-seGene <- transcriptToGeneExpression(se)
-seGene
+seGene.multiSample <- transcriptToGeneExpression(se.multiSample)
```
andredsim, line 254: Error in `:=`(TXNAME, "Incomaptible") : 
Check that is.data.table(DT) == TRUE. Otherwise, := and `:=`(...) are defined for use in j, once only and in particular ways. See help(":=").

vignettes/bambu.Rmd
```diff
@@ -204,25 +239,31 @@ individual genes. Here we look at gene ENSG00000099968 and visualise the
...
+seGene.multiSample
 ```
 
 We can again use the \code{plotBambu} function to visualise the gene expression
 data across the 6 samples with a heatmap or PCA plot. As expected, samples
 from the same cell line showed higher correlation than across the cell lines.
 ```{r, fig.width = 8, fig.height = 6}
-colData(seGene)$groupVar <- SGNexSamples$cellLine
-plotBambu(seGene, type = "heatmap")
+colData(seGene.multiSample)$groupVar <- SGNexSamples$cellLine
```
andredsim, line 262: couldn't test as I didn't have seGene.multiSample

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
 transcripts to text files.   
 ```{r}
 save.dir <- tempdir()
-writeBambuOutput(se, path = save.dir, prefix = "NanoporeRNASeq_")
+writeBambuOutput(se.multiSample, path = save.dir, prefix = "NanoporeRNASeq_")
```
andredsim, line 273: Error in `:=`(TXNAME, "Incomaptible") : 
Check that is.data.table(DT) == TRUE. Otherwise, := and `:=`(...) are defined for use in j, once only and in particular ways. See help(":=").

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
...
+
+Genome: Homo_sapiens.GRCh38.dna_sm.primary_assembly <br>
+Annotations: Homo_sapiens.GRCh38.91<br>
+Read count:  7,861,846<br>
+Technology: Nanopore (ONT)<br>
+Library preparation: directRNA<br>
+Base Calling Accuracy: 79%<br>
+Average Read Length: 1093<br>
+
+We have found the pretrained model works successfully across species borders (on Arabidopsis thaliana) and on different technologies (PacBio), with only small decreases in performance compared to using a sample specific model. The pretrained model is not always effective in samples with large differences in sequencing quality or if the library preparation results in biases in the overall structure of the transcriptome. In this case, we would recommend training a new model using similar data from a different sample that has quality reference annotations (See [Training a model on another species/dataset and applying it](#Training-a-model-on-another-speciesdataset-and-applying-it)).
```
andredsim:
```suggestion
We have found the pretrained model works successfully across species borders (on Arabidopsis thaliana) and on different technologies (PacBio), with only small decreases in performance compared to using a sample specific model. The pretrained model is not always effective in samples with large differences in sequencing quality or if the library preparation results in biases in the overall structure of the transcriptome. In this case, we would recommend training a new model using similar data from a different sample that has quality reference annotations.
```

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
...
 
+For the example in this vignette, the data used is rather small, hence the pre-trained model is used by default. So you won't see differences in the output, but you may try it out with more data (see [SG-NEx data](https://github.com/GoekeLab/sg-nex-data))
```
andredsim, line 303:
```suggestion
For the example in this vignette, the data used is rather small, hence the pre-trained model is used by default. So you won't see differences in the output, but you may try it out with more data
```

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
...
 
+### De-novo transcript discovery
 
+In cases where the organism does not yet have reference annotations, or unreliable annotations, *bambu* can be run in de-novo mode. In de-novo mode, *bambu* does not train a model, and instead uses the pretrained model to classify novel transcripts (see [Using a pretrained model](#Using-a-pretrained-model). To learn how to train a new model for a more closely related organism/sample see [Training a model on another species/dataset and applying it](#Training-a-model-on-another-speciesdataset-and-applying-it). Without annotations *bambu* is unable to calibrate the NDR output, nor be able to recommend a threshold and will instead use the TPS as the thresholded value. Therefore you should supply a manual NDR threshold ([Modulating the sensitivity of discovery (pre and post analysis)](#Modulating-the-sensitivity-of-discovery-pre-and-post-analysis)) and note that the precision of the output is unlikely to linearly match an applied threshold.
```
andredsim:
```suggestion
In cases where the organism does not yet have reference annotations, or unreliable annotations, *bambu* can be run in de-novo mode. In de-novo mode, *bambu* does not train a model, and instead uses the pretrained model to classify novel transcripts. Without annotations *bambu* is unable to calibrate the NDR output, nor be able to recommend a threshold and will instead use the TPS as the thresholded value. Therefore you should supply a manual NDR threshold and note that the precision of the output is unlikely to linearly match an applied threshold.
```

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
...
+The TPS threshold used is (> 1-NDR). If an NDR is not provided, a default NDR threshold of <0.1 is used (an effective TPS threshold of > 0.9). As in  [Modulating the sensitivity of discovery (pre and post analysis)](#Modulating-the-sensitivity-of-discovery-pre-and-post-analysis) an NDR of 1 can be provided to output all possible read classes with their TPS scores
```
andredsim:
```suggestion
The TPS threshold used is (> 1-NDR). If an NDR is not provided, a default NDR threshold of <0.1 is used (an effective TPS threshold of > 0.9). An NDR of 1 can be provided to output all possible read classes with their TPS scores 
```

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
...
-## Identifying differentially expressed genes {#DESeq2}
+```{r, results = 'hide'}
+novelAnnotations <- bambu(reads = bamFiles, annotations = NULL, genome = genomeSequence, NDR = 1, quant = FALSE)
+```
+### Storing and using preprocessed files (rcFiles)
+
+The first step of *bambu* involves the construction of read classes which is a large fraction of the running time. This could be time-consuming if we want to perform transcript discovery & quantification multiple times on the same dataset using different configurations (eg. NDR, or combinations of samples), especially when the sample is large. To mitigate this, we can store the read class information as read class files (rcFiles) during a *bambu* run. Then they can be used as an input argument in the  *bambu* main function for the subsequent *bambu* run.
+
+```{r, eval = FALSE} 
+se <- bambu(reads = rcFiles, annotations = annotations, genome = fa.file)
```
andredsim, line 318: should we maybe comment this out so people don't try and run it since it can't run without rcFiles?

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
...
+
+
+
+### Training a model on another species/dataset and applying it
+
+In situations where training is not or cannot be performed, and the default model is also not suitable for the sample (the sample is generated from a different technology, species, condition, etc), *bambu* provides the option to train a new model, if well annotated similar data is available. For example one might train a model on arabidopsis to apply to an unannotated plant sample.
+
+```{r, eval = FALSE}
+# first train the model using a related annotated dataset from .bam
+se = bambu(reads = sample1.bam, annotations = annotations, genome = fa.file, discovery = FALSE, quant = FALSE, opt.discovery = list(returnModel = TRUE)) # note that discovery and quant need to be set to FALSE, alternatively you can have them set to TRUE and retrieve the model from the rcFile as long as returnModel = TRUE ([see here](#Storing-and-using-preprocessed-files-rcFiles)).
```
andredsim, line 368: should we comment this out since they can't run this as part of the vignette?

vignettes/bambu.Rmd
```diff
@@ -232,29 +273,139 @@ and the gene expression estimates that include any newly discovered genes and
...
 *DESeq2* to find the differentially expressed genes between MCF7 and K562 cell
 lines. Similar to using results from *Salmon*, estimates from *bambu* will
 first be rounded.
+
+
 ```{r}
 library(DESeq2)
-dds <- DESeqDataSetFromMatrix(round(assays(seGene)$counts),
-                                    colData = colData(se),
+dds <- DESeqDataSetFromMatrix(round(assays(seGene.multiSample)$counts),
```
andredsim, line 404: I couldn't test this with seGene.multiSample but it did work with se.multiSample

andredsim (commented):
Looks good, here are a few suggestions!

vignettes/bambu.Rmd
cying111: can you try to install complexheatmap and try again? it runs through on my side without errors

vignettes/bambu.Rmd
cying111: for this one, when you use se.multiSample, does it still give you error?

vignettes/bambu.Rmd
cying111: same here, when you replace se with se.multiSample, does it still give you error?

PR #319, opened 2022-10-03
Address NDR update comments

andredsim:
Addresses the comments in https://github.com/GoekeLab/bambu/pull/267

Additionally makes the three following changes:
updated the default model so that the lm behind it can be used for predictions
ReadIds are no longer stored if trackReads is on.
Bug fix when no novel transcripts are detected


R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -124,34 +119,30 @@ filterTranscriptsByAnnotation <- function(rowDataCombined, annotationGrangesList
 }
 
 #' calculates an expected NDR based on the annotations'
-recommendNDR <- function(combinedTranscripts, baseline = 0.8, NDR = NULL){
-    message("-- Predicting annotation completeness to determine NDR threshold --")
+recommendNDR <- function(combinedTranscripts, baselineFDR = 0.1, NDR = NULL, defaultModels = defauleModels, verbose = FALSE){
```
jonathangoeke, line 122: typo
```suggestion
recommendNDR <- function(combinedTranscripts, baselineFDR = 0.1, NDR = NULL, defaultModels = defaultModels, verbose = FALSE){
```

PR #347, opened 2023-01-09
Bug FIxes

andredsim:
Covers severel niche crashes Bambu can encounter when certain read types are missing

Updates the warning system so warnings are output as they arise during verbose mode

### DIFF
--- a/R/bambu-processReads.R
+++ b/R/bambu-processReads.R
@@ -68,20 +68,25 @@ bambu.processReadsByFile <- function(bam.file, genomeSequence, annotations,
     readGrgList <- prepareDataFromBam(bam.file[[1]], verbose = verbose, use.names = trackReads)
     warnings = c()
     warnings = seqlevelCheckReadsAnnotation(readGrgList, annotations)
+    if(verbose & length(warnings) > 0) warning(paste(warnings,collapse = "\n"))
     #check seqlevels for consistency, drop ranges not present in genomeSequence
     refSeqLevels <- seqlevels(genomeSequence)
     if (!all(seqlevels(readGrgList) %in% refSeqLevels)) {
         refSeqLevels <- intersect(refSeqLevels, seqlevels(readGrgList))
         if (!all(seqlevels(annotations) %in% refSeqLevels)&(!(length(annotations)==0))) {
             refSeqLevels <- intersect(refSeqLevels, seqlevels(annotations))
-            warnings = c(warnings, paste0("not all chromosomes from annotations present in ", 
+            warning = paste0("not all chromosomes from annotations present in ", 
             "reference genome sequence, annotations without reference genomic sequence ",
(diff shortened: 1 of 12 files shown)


R/bambu-processReads.R
```diff
@@ -68,20 +68,25 @@ bambu.processReadsByFile <- function(bam.file, genomeSequence, annotations,
     readGrgList <- prepareDataFromBam(bam.file[[1]], verbose = verbose, use.names = trackReads)
     warnings = c()
```
cying111, line 69: this is not used?

R/bambu-processReads_scoreReadClasses.R
```diff
@@ -34,24 +40,30 @@ scoreReadClasses = function(se, genomeSequence, annotations, defaultModels,
...
 
     model = NULL
-    if (fit){ 
+    rowData(se)$txScore = rowData(se)$txScore.noFit
+    if (fit & length(thresholdIndex)>0){ 
         model = trainBambu(se, verbose = verbose)
         if(returnModel) metadata(se)$model = model
         txScore = getTranscriptScore(rowData(se)[thresholdIndex,], model,
                                  defaultModels)
+                                 print(5)
```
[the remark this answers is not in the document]
andredsim: nope, thanks for flagging it

R/bambu.R
```diff
@@ -153,7 +152,9 @@ bambu <- function(reads, annotations = NULL, genome = NULL, NDR = NULL,
 
     rm.readClassSe <- FALSE
     readClassList = reads
-    isBamFiles = ifelse(!is(reads, "BamFileList"), all(grepl(".bam$", reads)), FALSE)
+    isRDSs = all(sapply(reads, class)=="RangedSummarizedExperiment")
+    isBamFiles = !isRDSs
+    if(!isRDSs) isBamFiles = ifelse(!is(reads, "BamFileList"), all(grepl(".bam$", reads)), FALSE)
```
cying111, line 157: would the following be enough to replace 156-157?
```suggestion
     isBamFiles <- all(grepl(".bam$", reads))
```

cying111 (approved):

R/bambu.R
andredsim: This part of the code is very ugly, unfortunately the problem is using all() on a rangedsummarizedexperiment is very very slow which is why I check first that it isn't the loaded in .rds files.

jonathangoeke: is it possible to provide readclasses directly? I thought it only works with bam files or a list of read class files? (not with read class objects in memory)

jonathangoeke: (I guess the answer is that it is possible...looking through the next code, which is probably helpful to have anyway)

jonathangoeke (approved):

Issue #353, opened 2023-02-02
Issue with `writeBambuOutput`


paulimer:
Hi all,
With the test code of the README, followed by the line
```r
writeBambuOutput(se, "path/to/out_dir")
```
produces the error 
```r
Error in `left_join()`:
! `...` must be empty.
✖ Problematic argument:
• all = TRUE
```
which seems to be produced by [this line](https://github.com/GoekeLab/bambu/blob/master/R/readWrite.R#L99). 
From what I've seen the `all = TRUE` argument is a valid `dplyr::left_join` argument. Perhaps it was the case in a previous dplyr version?
Thanks in advance!

gabee-chan:
I'm also having this problem. :( Does anyone know a solution to get the writeBambuOutput() outputs?

cying111:
Hi @paulimer @gabee-chan , thanks for pointing this problem out! It turns out this error was related to the update of package dplyr, there are two solutions to this:
1) either use the older version of dplyr
2) use the [bug fix branch](https://github.com/GoekeLab/bambu/tree/writeBambuOutput_patch) for now
 
while we are updating the release branch

Please do try out the solutions and let us know if the error persists

I will also come back again to this issue once we have updated the release branch

cying111:
Hi here, the bug is now fixed in the release branch, I will close this issue now. Thank you!

PR #355, opened 2023-02-08
bug fix 

cying111:
- related to [issue 353](https://github.com/GoekeLab/bambu/issues/353) , this is related to the update of dplyr package, where all = TRUE argument is no longer valid in left_join, also the bug fix for  the build error of the release package
- factor in incompatible counts for CPM calculation (previously didn't)


R/bambu-quantify.R
```diff
@@ -24,14 +24,16 @@ bambu.quantify <- function(readClass, annotations, emParameters,
     incompatibleCounts <- incompatibleCounts[data.table(GENEID = unique(mcols(annotations)$GENEID)), on = "GENEID"]
     incompatibleCounts[is.na(counts), counts := 0]
     setnames(incompatibleCounts, "counts", colnames(readClass))
+    compatibleCounts <- calculateCPM(compatibleCounts, incompatibleCounts)
```
andredsim, line 27: at this point in the code, incompatibleCounts$counts doesn't exist because of line 26. The column is labeled by the sample name so the sum($counts) is 0. This would change the calculation I think?
This means the sum of $CPM is 1000000 even though the existance of incompatibleCounts should make this lower

andredsim (commented):
Just found the one potential error in calculateCPM(), could be that I misunderstand the intention though.

R/bambu-quantify.R
cying111: thanks for pointing this out! fixed it now

R/bambu-quantify_utilityFunctions.R
```diff
@@ -454,16 +454,16 @@ modifyQuantOut <- function(outEst, outIni){
...
-    theta_est[, `:=`(CPM = counts / totalCount * (10^6))]
-    return(theta_est)
-}
+# formatOutput <- function(theta_est){
+#     theta_est <- theta_est[, .(txid, counts, fullLengthCounts,
+#     uniqueCounts)]
+#     # totalCount <- sum(theta_est$counts)
+#     # theta_est[, `:=`(CPM = counts / totalCount * (10^6))]
+#     return(theta_est)
+# }
```
jonathangoeke, line 466: remove this?

R/bambu-quantify.R
```diff
@@ -23,15 +23,17 @@ bambu.quantify <- function(readClass, annotations, emParameters,
                                ncore = ncore, verbose = verbose)
     incompatibleCounts <- incompatibleCounts[data.table(GENEID = unique(mcols(annotations)$GENEID)), on = "GENEID"]
     incompatibleCounts[is.na(counts), counts := 0]
+    compatibleCounts <- calculateCPM(compatibleCounts, incompatibleCounts)
     setnames(incompatibleCounts, "counts", colnames(readClass))
```
jonathangoeke, line 27: I'm not sure why this line is needed here? incompatableCounts is not used anymore?

jonathangoeke (approved):

R/bambu-quantify_utilityFunctions.R
cying111: sure, i will remove this.

R/bambu-quantify.R
cying111: This line is to calculate CPM with incompatible counts used. The incompatible counts will still be used later

andredsim (approved):

Issue #364, opened 2023-04-19
Missing value error from checkFeatures


ChangqingW:
I am getting a missing value error during isoform discovery:
```
Model to predict true splice sites built in 0 mins.
reads count for all annotated junctions: 29662 (0.891366409231601%)
reads count for all annotated junctions after correction to reference junction: 19176 (0.576253869038675%)
Finished correcting junction based on set of high confidence junctions in 0 mins.
Finished creating transcript models (read classes) for reads with spliced junctions in 0 mins.
Finished create single exon transcript models (read classes) in 0 mins.
Error: BiocParallel errors
  1 remote errors, element index: 1
  0 unevaluated and other errors
  first remote error:
Error in if (sum(labels) == length(labels) | sum(labels) == 0) {: missing value where TRUE/FALSE needed
In addition: Warning messages:
1: In bambu.processReadsByFile(bam.file = reads[bamFileName], genomeSequence = genomeSequence,  :
  not all chromosomes present in reference annotations, annotations might be incomplete. Please compare objects on the same reference
2: In bambu.processReadsByFile(bam.file = reads[bamFileName], genomeSequence = genomeSequence,  :
  1 reads are mapped outside the provided genomic regions. These reads will be dropped. Check you are using the same genome used for the alignment
```
Which seems to be from [checkFeatures](https://github.com/GoekeLab/bambu/blob/5c758f5d7c2b8a899143b4526985c6b84d4bfb7f/R/bambu-processReads_scoreReadClasses.R#L303).   
I am using GRCh38 and gencode 33.
Would really appreciate it if you could help me on solving this issue.

ChangqingW:
```
> .traceback()
[[1]]
[1] "stop(.error_bplist(res))"

[[2]]
[1] ".bpinit(manager = manager, X = X, FUN = FUN, ARGS = ARGS, BPPARAM = BPPARAM, "
[2] "    BPOPTIONS = BPOPTIONS, BPREDO = BPREDO)"

[[3]]
[1] "bplapply(names(reads), function(bamFileName) {"
[2] "    bambu.processReadsByFile(bam.file = reads[bamFileName], genomeSequence = genomeSequence, "
[3] "        annotations = annotations, readClass.outputDir = readClass.outputDir, "
[4] "        stranded = stranded, min.readCount = min.readCount, fitReadClassModel = fitReadClassModel, "
[5] "        min.exonOverlap = min.exonOverlap, defaultModels = defaultModels, "
[...]

PR #366, opened 2023-04-27
Small bug fix for Issue #364

andredsim:
Bug was happening because when a higher min.readCount was set, read classes below this threshold would still be passed in the model training but because of the threshold, they had not been assigned appropriate labels for training.


R/bambu-processReads_scoreReadClasses.R
```diff
@@ -302,6 +302,7 @@ prepareTranscriptModelFeatures = function(rowData){
 checkFeatures = function(features, verbose = FALSE){
     labels = features$labels
     trainable = TRUE
+    print(features)
```
jonathangoeke: is this required?

jonathangoeke (approved):

cying111 (approved):

PR #406, opened 2023-12-21
fixing the chunk when showing the import from the TxDb object

federicomarini:
I simply changed the vignette chunk to have it really reading in the TxDb file first and then `loadDb`ing it for using as input to `prepareAnnotations()`


cying111 (approved):

jonathangoeke (approved):
thanks for finding this

federicomarini:
Glad I could contribute a microbit to this beauty ❤️

PR #452, opened 2024-10-25
Multiplex_Major_Patch

andredsim:
**Work in progress. Do not use or run**

From NDR PR

In this PR transcripts that are normally filtered out including subset transcripts and those above the NDR threshold are placed into the metadata of the annotations in $subsetTranscripts and $lowConfidenceTranscripts respectively.
This means that if users ran bambu with the wrong NDR setting, and do not want to run discovery again, they can get the missing transcripts from the metadata.
To facilitate this this PR adds the external setNDR function which takes the extendedAnnotations and an NDR value and will switch novel bambu annotations between the main annotations and the low confidence annotations based on the threshold provided. If no threshold is provided, setNDR will recommend an NDR with the same method used during transcript discovery.
In order for this to work for annotations that have already been saved to a gtf file, bambu now outputs the NDR, txScore and txScore.noFit as attributes to the gtf file and these are also read in with prepareAnnotations.
Important to note that if annotations are written with an NDR threshold of <1, these low confidence transcripts will be missed.
Added setNDR as part of quant, which means that users can provide their extendedAnnotations alongside an NDR threshold when running bambu and it will automatically adjust the NDR used for quant. This means users do not need to manually filter the NDR value themselves.
NDR and other stats are now copied over to equal transcripts even if above the NDR threshold (previously only happened for those below the NDR threshold)
Minor change:
Warnings will no longer occur if there are seqlevels in the readGrgList that are not in the annotations or genome. This was done by setting seqlevels of the reads to only those in the reads. Warning was constantly occuring because all the scaffolds used in alignment were in the bam files, even if no reads from these scaffolds existed.

For SC PR

1) major changes that impact results
Updated the xgboost models to the latest version

2) major changes that change the user interface
Deprecated lowMemory - This has been replaced by processByChromosome
Added processByChomosome (the old memory)
Added processByBam argument
writeBambuOutput now outputs all information needed to import Bambu results from text files
Count outputs are all now in sparse matrix format

3) major changes that introduce new options like the single cell option
Added demultiplexed argument
Added spatial argument
Added sampleNames argument
Added cleanReads argument
Added dedupUMI argument
Added clusters argument

4) minor changes
Added importBambuResults


R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -19,14 +19,15 @@ isore.combineTranscriptCandidates <- function(readClassList,
...
     combinedSplicedTranscripts[,confidenceType := "highConfidenceJunctionReads"]
-    combinedUnsplicedTranscripts <- 
-        combineUnsplicedTranscriptModels(readClassList, bpParameters, 
-        stranded, min.readCount, min.readFractionByGene, 
-        min.txScore.multiExon, min.txScore.singleExon, verbose) %>% data.table()
-    combinedUnsplicedTranscripts[, confidenceType := "unsplicedNew"]
-    combinedTranscripts <- as_tibble(rbindlist(list(combinedSplicedTranscripts,
-        combinedUnsplicedTranscripts), fill = TRUE))
-    return(combinedTranscripts)
+    return(combinedSplicedTranscripts)
```
cying111: Hi @andredsim 

where is the commented code moved to? does this mean there is no unsplicedNew transcript models?

andredsim: This is an outdated version, just double check you have pulled the latest version of this branch, as the commented code is uncommented in it. Now the unsplicedNew transcript models only appear if the user provides min.txScore.singleExon < 1. Which is essentially the same as before as we didn't output novel single exon transcripts by default.

R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -80,25 +84,36 @@ filterTranscriptsByAnnotation <- function(rowDataCombined, annotationGrangesList
...
   rowDataCombined = calculateNDROnTranscripts(rowDataCombined, 
                         useTxScore = length(annotationGrangesList)==0)
-
   if(length(annotationGrangesList)>0){ #only recommend an NDR if its possible to calculate an NDR
       NDR = recommendNDR(rowDataCombined, baselineFDR, NDR, defaultModels, verbose)
   } else {
       if(is.null(NDR)) NDR = 0.5
   }
-  filterSet = (rowDataCombined$NDR <= NDR)
+  filterSet = (rowDataCombined$NDR <= NDR | rowDataCombined$readClassType == "equal:compatible")
```
cying111: why the equal read class needed here? would they just be the annotated transcripts?

R/bambu.R
```diff
@@ -136,24 +136,48 @@
 #'     genome = fa.file,  discovery = TRUE, quant = TRUE)
 #' @export
 bambu <- function(reads, annotations = NULL, genome = NULL, NDR = NULL,
-    opt.discovery = NULL, opt.em = NULL, rcOutDir = NULL, discovery = TRUE, 
-    quant = TRUE, stranded = FALSE,  ncore = 1, yieldSize = NULL,  
+    mode = NULL, opt.discovery = NULL, opt.em = NULL, rcOutDir = NULL, discovery = TRUE,
```
cying111, line 139: the mode parameter is very helpful, would be great to have pre-defined parameters for each mode documented in the readme or documentation

R/bambu.R
```diff
@@ -136,24 +136,48 @@
...
-    fusionMode = FALSE, verbose = FALSE) {
+    fusionMode = FALSE, verbose = FALSE, demultiplexed = FALSE, spatial = NULL, quantData = NULL,
+    sampleNames = NULL, cleanReads = FALSE, dedupUMI = FALSE, barcodesToFilter = NULL, clusters = NULL) {
+    message(paste0("Running Bambu-v", "3.3.0"))
+    if(!is.null(mode)){
+        if(mode == "bulk"){
+            lowMemory = TRUE
+        }
+        if(mode == "multiplexed"){
+            if(is.null(demultiplex)) demultiplex = TRUE
```
cying111:
```suggestion
            if(is.null(demultiplex)) demultiplexed = TRUE
```

cying111: is it better to default it to FALSE when mode is "multiplexed"?
then give option for providing demultiplexed bam file

R/bambu.R
```diff
@@ -136,24 +136,48 @@
...
 bambu <- function(reads, annotations = NULL, genome = NULL, NDR = NULL,
-    opt.discovery = NULL, opt.em = NULL, rcOutDir = NULL, discovery = TRUE, 
-    quant = TRUE, stranded = FALSE,  ncore = 1, yieldSize = NULL,  
+    mode = NULL, opt.discovery = NULL, opt.em = NULL, rcOutDir = NULL, discovery = TRUE, 
+    assignDist = TRUE, quant = TRUE, stranded = FALSE,  ncore = 1, yieldSize = NULL,  
     trackReads = FALSE, returnDistTable = FALSE, lowMemory = FALSE, 
-    fusionMode = FALSE, verbose = FALSE) {
+    fusionMode = FALSE, verbose = FALSE, demultiplexed = FALSE, spatial = NULL, quantData = NULL,
+    sampleNames = NULL, cleanReads = FALSE, dedupUMI = FALSE, barcodesToFilter = NULL, clusters = NULL) {
+    message(paste0("Running Bambu-v", "3.3.0"))
```
cying111:
```suggestion
    message(packageVersion("bambu")) 
```

cying111: not sure why this is needed though

PR #460, opened 2025-01-27
Remove the prefix of gtf file

### LINKED ISSUE #464: Document the changes of write and import bambu result
1. Remove “extended_annotations” in output gtf file name.
2. Add “_” between prefix and file name.
3. Output an empty gtf file when where is no novel transcript.
4. Output barcode.tsv when the input the single-cell data.

SuiYue-2308:
Remove the prefix "extended_annotations" of output gtf files

### DIFF
--- a/R/readWrite.R
+++ b/R/readWrite.R
@@ -173,7 +173,7 @@ writeToGTF <- function(annotation, file, geneIDs = NULL) {
         frame = ".", attributes = paste(GENEID, group_name, exon_rank, NDR, txScore, txScore.noFit, novelGene, novelTranscript, txClassDescription )) %>%
         select(seqnames, source, feature, start, end, score,
         strand, frame, attributes, group_name)
-    dfTx <- as.data.frame(range(ranges(annotation)))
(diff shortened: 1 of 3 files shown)


R/readWrite.R
```diff
@@ -27,8 +27,7 @@ writeBambuOutput <- function(se, path, prefix = "", outputExtendedAnno = TRUE,
             dir.create(outdir, recursive = TRUE)
 
         transcript_grList <- rowRanges(se)
-        transcript_gtffn <- paste(outdir, prefix,
-            "extended_annotations", sep = "")
+        transcript_gtffn <- paste(outdir, prefix, sep = "")
```
cying111, line 31: Currently prefix is assumed to be NULL, in the case of prefix is not NULL, are we going to have a underscore in between prefix and the names following it? 
```suggestion
        transcript_gtffn <- ifelse(is.null(prefix), outDir, paste(outdir, prefix, "_",sep = ""))
```

cying111 (approved):
Looks good to me!

Issue #450, opened 2024-10-17
Large numebr of BAM files leads to Error in `vec_interleave_indices()`:


NikoLichi:
Dear Bambu team,

I am running a massive project with 480 BAM files with ~4.8 TB total data.
Following the previous suggestion for Bambu, I am running first the extended annotations (`quant = FALSE`), with the idea of running the quantification later in batches.

However, the is a major issue when starting the extended annotations: 
```
--- Start extending annotations ---
Error in `vec_interleave_indices()`:
! Long vectors are not yet supported in `vec_interleave()`. Result from interleaving would have size 8857886400, which is larger than the maximum supported size of 2^31 - 1.
Backtrace:
     ▆
  1. ├─bambu::bambu(...)
  2. │ └─bambu:::bambu.extendAnnotations(...)
  3. │   └─bambu:::isore.combineTranscriptCandidates(...)
  4. │     ├─... %>% data.table()
  5. │     └─bambu:::combineSplicedTranscriptModels(...)
  6. │       └─bambu:::updateStartEndReadCount(combinedFeatureTibble)
  7. │         └─... %>% mutate(sumReadCount = sum(readCount, na.rm = TRUE))
  8. ├─data.table::data.table(.)
  9. ├─dplyr::mutate(., sumReadCount = sum(readCount, na.rm = TRUE))
 10. ├─dplyr::group_by(., rowID)
 11. ├─tidyr::pivot_longer(...)
 12. ├─tidyr:::pivot_longer.data.frame(...)
 13. │ └─tidyr::pivot_longer_spec(...)
 14. │   └─vctrs::vec_interleave(!!!val_cols, .ptype = val_type)
 15. │     └─vctrs:::vec_interleave_indices(n, size)
 16. └─rlang::abort(message = message)
Execution halted
```

Is there anything I could do to run Bambu?

My code looks like:
```
BAMlist = BAMs_one_per_Line
fa.file <- "/refData/release46/GRCh38.primary_assembly.genome.fa"
gtf.file <-  "/refData/release46/gencode.v46.primary_assembly.annotation.gtf"
bambuAnnotations <- prepareAnnotations(gtf.file)

extendedAnnotations = bambu(reads = BAMlist, annotations = bambuAnnotations, genome = fa.file, quant = FALSE, lowMemory=T, ncore = 14, rcOutDir="MY_PATH/bambu_20241015_all")

```

As an additional note, I also have the same warning message as some others have reported as issue #407 .

[...]

PR #462, opened 2025-02-03
Patch bigsamples

cying111:
- fix for [issue](https://github.com/GoekeLab/bambu/issues/450)

Brief description of the issue:
the issue is caused by when processing RC start/end from different samples, there is a step to find the weighted median across different samples, which involves explicitly write out each sample start/end as a column, and then convert the dataframe from wide to long format.

The use of pivot_longer() is crashed when sample size is big and RC number is also big, i.e., the matrix dimension is huge.

Solution:
To solve that, I avoided using pivot_longer and any format of wide to long conversion. Instead, I used a function directly on the columns, to get the one value for each RC. By doing so, it avoided expanding the data, which caused R crash.


jonathangoeke (approved):

SuiYue-2308 (approved):
bigsample returns the same se object with devel

PR #470, opened 2025-02-27
remove the lowMemory and update example code for generating rcf

SuiYue-2308:
lowMemory causes error when specifying rcOutDir
update examples in readme about generate rcf (assignDist need to be FALSE when generating rcf)

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,9 +330,9 @@ se <- bambu(reads = rcFiles, annotations = annotations, genome = fa.file)
 ```
 rcFiles can be generated in two ways, either as a direct output of the bambu() function when quant and discovery are FALSE, or as written outputs when a path is provided to the rcOutdir argument. When rcFiles are output using rcOutdir this is done using BiocFileCache. For more details on how to access, use, and identify these files see [here](https://bioconductor.org/packages/release/bioc/html/BiocFileCache.html). A short example is shown below.
 
-Example using rcOutDir to produce preprocessed files, and se returns to the path of files
+Example using rcOutDir to produce preprocessed files
 ```rscript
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -330,9 +330,9 @@ se <- bambu(reads = rcFiles, annotations = annotations, genome = fa.file)
 ```
 rcFiles can be generated in two ways, either as a direct output of the bambu() function when quant and discovery are FALSE, or as written outputs when a path is provided to the rcOutdir argument. When rcFiles are output using rcOutdir this is done using BiocFileCache. For more details on how to access, use, and identify these files see [here](https://bioconductor.org/packages/release/bioc/html/BiocFileCache.html). A short example is shown below.
 
-Example using rcOutDir to produce preprocessed files
+Example using rcOutDir to produce preprocessed files, and se returns to the path of files
 ```rscript
-se <- bambu(reads = test.bam, rcOutDir = "path/to/rcOutput/", annotations = annotations, genome = fa.file)
+se <- bambu(reads = test.bam, rcOutDir = "path/to/rcOutput/", annotations = annotations, genome = fa.file, assignDist = FALSE)
```
cying111: why do you change this line and the description above? I think this example is more to demonstrate how to save rcfile? so the original description and command makes sense still

SuiYue-2308: Hi Chen Ying, because without the assignDist = FALSE, se returns to the transcripts, not the read class and include the annotations. And about the description, I would like to claim that se is path of rcf not read class object. If this looks strange then I can also change it back.

SuiYue-2308: Hi Chen Ying, I have removed this part of change. Thank you so much for you help!

cying111 (approved):

PR #501, opened 2025-06-27
Fix bugs in writeBambuOutput. 

SuiYue-2308:
Correct the order of incompatibleCounts.mtx to match to gene.tsv.

### DIFF
--- a/R/readWrite.R
+++ b/R/readWrite.R
@@ -40,14 +40,14 @@ writeBambuOutput <- function(se, path, prefix = "", outputExtendedAnno = TRUE,
                             feature='transcript',outdir, prefix)
         print(d)
         }
-        seGene <- transcriptToGeneExpression(se)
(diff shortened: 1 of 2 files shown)


lingminhao:
@SuiYue-2308 , I suggest to fix the order of GENEID directly within the `transcriptToGeneExpression(se)` instead, then you do not have to reorder `metadata(se)$incompatibleCounts` anymore.

SuiYue-2308:
> @SuiYue-2308 , I suggest to fix the order of GENEID directly within the `transcriptToGeneExpression(se)` instead, then you do not have to reorder `metadata(se)$incompatibleCounts` anymore.

Hi @lingminhao , thank you for the suggestion. I think this is a better way with minimal code changes. I’ve tested the results after your commit, and everything looks good to me. Thanks!

lingminhao (approved):

cying111 (approved):

PR #512, opened 2025-12-01
Fix bam format

ch99l:
Fixed syntax to allow bambu to process bam files generated from Epi2Me and IsoSeq


SuiYue-2308 (approved):
Hi, I’ve reviewed the code and it looks good to me.
I also tested it on a PacBio single-cell BAM file from IsoSeq and didn’t encounter any issues. It should be safe to combine.

lingminhao (approved):
Code change ok. Description - Standardise syntax for cell barcode and umi (as commonly used by others)

PR #516, opened 2026-01-07
Xgboost bug fix

cying111:
This PR is created to resolve the issues related to xgboost updates, specifically:
- update all label features to be type logical in both testSpliceSite and getTranscriptScores function
- update pre-saved models for both splicing junction prediction and defaultModels for read class scoring
- update test code for xgboost as well as saved outputs relevant to xgboost testing


SuiYue-2308 (approved):
Hi, I’ve reviewed the code changes and everything looks good. I also tested it on a real dataset (dRNA), and it ran smoothly. 

Update about xgboost pull request.
- In annotation-free mode, the se objects produced by the two branches are identical.
- Annotation-provided mode, the se objects are no longer identical.
    (1) NDR score is similar (0.04597701 v.s. 0.04595099)
    (2) Novel transcripts: The number of novel transcripts differs between XGBoost versions (340 with XGBoost 1.7.11.1 vs. 345 with XGBoost 3.1.3.1). Among these, 279 transcripts have identical rowRanges.
    (3) Expression levels: Transcript-level read counts are largely similar; the differences might from XGBoost update.

data-raw/DATASET.R
```diff
@@ -102,24 +102,33 @@ seCombinedGeneExpected <- transcriptToGeneExpression(seCombined)
 seCombinedExtendedGeneExpected <- transcriptToGeneExpression(seCombinedExtended)
 
 
+defaultModels$transcriptModelME = NULL
+defaultModels$transcriptModelSE = NULL
+saveRDS(defaultModels, "./inst/extdata/defaultModels.rds")
```
jonathangoeke, line 107: why is this needed?

cying111: defaultModels is an object created by Andre, where there are 2 model components and some regular data components. This step is just to save the regular list component of the final defaultModels as rds while the model component as binary files (previously .model file, now .ubj file). The files are saved so that in case sysdata are changed accidentally, we can still get it back by calling these files.

jonathangoeke (approved):
please add release note. Thanks!

PR #517, opened 2026-01-08
Switching from summarise() to reframe() to avoid dplyr warnings for v4. 

SuiYue-2308:
This pull is for fix the duplicate single-exon transcripts in final output and also the dplyr warning.

### DIFF
--- a/R/bambu-extendAnnotations-utilityCombine.R
+++ b/R/bambu-extendAnnotations-utilityCombine.R
@@ -304,23 +304,14 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
         } , BPPARAM = bpParameters))
     newUnsplicedTibble <- newUnsplicedTibble %>% 
         left_join(rowDataCombined, by =  "row_id") %>%
-        separate(row_id, c("sample","rcName"), sep = "\\-") %>%
-        mutate(sample_id = as.integer(gsub("s","",sample))) %>%
-        mutate(sample_name = colDataNames[sample_id]) %>%
-        select(-sample, -sample_id) %>%
         mutate(readCount_tmp = readCount) %>%
-        group_by(chr,strand, start, end, sample_name) %>%
+        group_by(chr,strand, start, end) %>%
         summarise(readCount = sum(readCount),
-                    geneReadProp = sum(geneReadProp),
-                    txScore = weighted.mean(txScore, readCount_tmp),
-                    txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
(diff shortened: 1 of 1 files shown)


cying111 (approved):
Have tested on test data, also tested the function specifically using example data from function. Results are identical.

So looks good to me.

R/bambu-processReads_utilityConstructReadClasses.R
```diff
@@ -305,8 +305,7 @@ getUnsplicedReadClassByReference <- function(granges, grangesReference,
     ## and maximum start of all overlapping exons (identical to
     ## minimum equivalent class)
     hitsDF <- hitsDF %>% dplyr::select(queryHits, chr, start, end, strand) %>%
-      group_by(queryHits, chr, strand) %>%
-      summarise(start = max(start), end = min(end), .groups = "drop") %>%
+      reframe(.by = c(queryHits, chr, strand), start = max(start), end = min(end)) %>% ungroup() %>%
```
lingminhao: no need to ungroup() after reframe()

lingminhao: By default, reframe() gives ungrouped data frame, so ungroup() is redundant (for example above). Other than that, the change looks good to me.

lingminhao (approved):
By default, reframe() gives ungrouped data frame, so ungroup() is redundant (for example above). Other than that, the change looks good to me.

R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -309,13 +309,13 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
...
-        summarise(readCount = sum(readCount),
+        reframe(.by = c(chr,strand, start, end, sample_name),
+                    readCount = sum(readCount),
                     geneReadProp = sum(geneReadProp),
                     txScore = weighted.mean(txScore, readCount_tmp),
-                    txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
-        group_by(chr, strand, start, end) %>% 
-        summarise(readCount = sum(readCount),
+                    txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>% 
+        reframe(.by = c(chr,strand, start, end),
```
cying111: see my comment for the other PR, I think the same thing applies here, i.e. only this summarise usage need to be changed, but changing to mutate would suffice, I think.

The rest summarise usage no need to change.

cying111: I have tested on my side, that if just change summarise to mutate for this case, but others all kept unchanged, results are identical, on a big sample.

Issue #502, opened 2025-07-21
summarise in dplyr is depricated


dgreen42:
Hi,

The summarise() function in dplyr is deprecated:

2: Returning more (or less) than 1 row per `summarise()` group was deprecated in dplyr 1.1.0.
ℹ Please use `reframe()` instead.
ℹ When switching from `summarise()` to `reframe()`, remember that `reframe()` always returns an
  ungrouped data frame and adjust accordingly.
ℹ The deprecated feature was likely used in the bambu package.
  Please report the issue to the authors.

lingminhao:
Hi @dgreen42 , thanks for the reply. We are aware with this issue and we will update this to `reframe()` during the next release of bambu.

cying111:
Hi @dgreen42 ,

The bug fix should be reflected in the most recent Bambu version 3.12.1 now. 

Please feel free to give it a try and let us know if same issue remain.

For now, I will close this issue.

Thank you

PR #518, opened 2026-01-08
Switching from summarise() to reframe() to avoid dplyr warnings

### LINKED ISSUE #408: dplyr::summarise message
The following messages appears at the end of a (successful) run of bambu, it's a minor warning from dplyr:

```
...
--- Start extending annotations ---
 Using a novel discovery rate (NDR) of: 0.063
 --- Start isoform quantification ---
 --- Finished running Bambu ---
 Warning messages:
 1: There was 1 warning in `summarise()`.
 ℹ In argument: `intersectWidth = max(intersectWidth)`.
 Caused by warning in `max()`:
 ! no non-missing arguments to max; returning -Inf
 2: Returning more (or less) than 1 row per `summarise()` group was deprecated in dplyr 1.1.0.
 ℹ Please use `reframe()` instead.
 ℹ When switching from `summarise()` to `reframe()`, remember that `reframe()` always returns an ungrouped data frame and adjust accordingly.
 ℹ The deprecated feature was likely used in the bambu package.
   Please report the issue to the authors.
 This warning is displayed once every 8 hours.
 Call `lifecycle::last_lifecycle_warnings()` to see where this warning was generated.
 3: There was 1 warning in `summarise()`.
 ℹ In argument: `intersectWidth = max(intersectWidth)`.
 Caused by warning in `max()`:
 ! no non-missing arguments to max; returning -Inf
```

This is with R 4.3.2 and Bioc 3.18 / bambu (3.4.0)

SuiYue-2308:
This pull is for fix the duplicate single-exon transcripts in final output and also the dplyr warning.

1. The bug is actually we will have duplicate single-exon transcripts in the final se object. This occurred because read classes were being grouped by sample_name, preventing the merging of identical transcripts found across different samples.

1. I made these 2 changes
- Removed group_by(sample_name) to allow for global aggregation of read classes.
- Switched to readCount_tmp in the NSampleReadCount calculation, make sure we use the original per-sample read counts when validating read class against the min.readCount threshold

### DIFF
--- a/R/bambu-extendAnnotations-utilityCombine.R
+++ b/R/bambu-extendAnnotations-utilityCombine.R
@@ -300,23 +300,15 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
         } , BPPARAM = bpParameters))
     newUnsplicedTibble <- newUnsplicedTibble %>% 
         left_join(rowDataCombined, by =  "row_id") %>%
-        separate(row_id, c("sample","rcName"), sep = "\\-") %>%
-        mutate(sample_id = as.integer(gsub("s","",sample))) %>%
-        mutate(sample_name = colDataNames[sample_id]) %>%
-        select(-sample, -sample_id) %>%
         mutate(readCount_tmp = readCount) %>%
-        group_by(chr,strand, start, end, sample_name) %>%
+        group_by(chr,strand, start, end) %>%
         summarise(readCount = sum(readCount),
-                    geneReadProp = sum(geneReadProp),
-                    txScore = weighted.mean(txScore, readCount_tmp),
-                    txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
-        group_by(chr, strand, start, end) %>% 
-        mutate(readCount = sum(readCount),
-                    maxTxScore = txScore,
-                    maxTxScore.noFit = txScore.noFit,
-                    NSampleReadCount = sum(readCount >= min.readCount), 
-                    NSampleReadProp = sum(geneReadProp >= 
-                                            min.readFractionByGene),
-                    NSampleTxScore = sum(maxTxScore > min.txScore.singleExon))
+                  maxTxScore = weighted.mean(txScore, readCount_tmp),
+                  maxTxScore.noFit = weighted.mean(txScore.noFit, readCount_tmp),
+                  NSampleReadCount = sum(readCount_tmp >= min.readCount), 
+                  NSampleReadProp = sum(geneReadProp >= 
+                                          min.readFractionByGene),
+                  NSampleTxScore = sum(txScore > min.txScore.singleExon))
+    
     return(newUnsplicedTibble)
 }
--- a/R/bambu-extendAnnotations-utilityCombine.R
+++ b/R/bambu-extendAnnotations-utilityCombine.R
@@ -311,7 +311,7 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
                     txScore = weighted.mean(txScore, readCount_tmp),
                     txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
         group_by(chr, strand, start, end) %>% 
-        summarise(readCount = sum(readCount),
+        mutate(readCount = sum(readCount),
                     maxTxScore = txScore,
                     maxTxScore.noFit = txScore.noFit,
                     NSampleReadCount = sum(readCount >= min.readCount), 


cying111 (approved):
tested on test data, looks all good to me.

R/bambu-processReads_utilityConstructReadClasses.R
```diff
@@ -301,8 +301,7 @@ getUnsplicedReadClassByReference <- function(granges, grangesReference,
     ## and maximum start of all overlapping exons (identical to
     ## minimum equivalent class)
     hitsDF <- hitsDF %>% dplyr::select(queryHits, chr, start, end, strand) %>%
-      group_by(queryHits, chr, strand) %>%
-      summarise(start = max(start), end = min(end), .groups = "drop") %>%
+      reframe(.by = c(queryHits, chr, strand), start = max(start), end = min(end)) %>% ungroup() %>%
```
lingminhao: no need ungroup() after reframe()

lingminhao: By default, reframe() gives ungrouped data frame, so ungroup() is redundant (for example above). Other than that, the change looks good to me.

lingminhao (approved):
reframe() gives ungrouped data frame, so ungroup() is redundant (for example above). Other than that, the change looks good to me. 

It resolved the github issue #502

R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -305,13 +304,13 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
...
-        group_by(chr,strand, start, end, sample_name) %>%
-        summarise(readCount = sum(readCount),
+        reframe(.by = c(chr, strand, start, end, sample_name), 
+                    readCount = sum(readCount),
                     geneReadProp = sum(geneReadProp),
                     txScore = weighted.mean(txScore, readCount_tmp),
                     txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
-        group_by(chr, strand, start, end) %>% 
-        summarise(readCount = sum(readCount),
+        reframe(.by = c(chr, strand, start, end),
```
cying111: I think the warning really is coming from this usage only, where txScore is from multiple samples, hence, multiple rows will be produced in this step.

A quick fix would be change it to mutate, cause there is no summarising happening at this step.

All the other instances of summarise no need to change actually. I have tested on my side for this. 

Thanks

R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -311,7 +311,7 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
                     txScore = weighted.mean(txScore, readCount_tmp),
```
jonathangoeke:
```suggestion
                    maxTxScore = weighted.mean(txScore, readCount_tmp),
```

R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -311,7 +311,7 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
...
                     txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
```
jonathangoeke:
```suggestion
                    maxTxScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
```

R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -311,7 +311,7 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
...
         group_by(chr, strand, start, end) %>% 
-        summarise(readCount = sum(readCount),
+        mutate(readCount = sum(readCount),
                     maxTxScore = txScore,
```
jonathangoeke:
```suggestion
```

R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -311,7 +311,7 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
...
                     maxTxScore.noFit = txScore.noFit,
```
jonathangoeke:
```suggestion
```

R/bambu-extendAnnotations-utilityCombine.R
```diff
@@ -311,7 +311,7 @@ makeUnsplicedTibble <- function(combinedNewUnsplicedSe,newUnsplicedSeList,
                     txScore = weighted.mean(txScore, readCount_tmp),
                     txScore.noFit = weighted.mean(txScore.noFit, readCount_tmp)) %>%
         group_by(chr, strand, start, end) %>%
```
jonathangoeke:
```suggestion
       select(chr, strand, start, end,readCount,maxTxScore,maxTxScore.noFit,geneReadProp) %>%
       group_by(chr, strand, start, end) %>% 
```

PR #520, opened 2026-01-09
fix xgboost data and label rename to x and y warning

cying111:
Fix the warning related to xgboost argument updates from data and label to x and y


SuiYue-2308 (approved):

jonathangoeke (approved):

jonathangoeke (commented):
can you add a note for the release here?

PR #521, opened 2026-01-09
fix xgboost data and label rename to x and y warning

cying111:
Fix the warning related to xgboost argument renaming from data and label to x and y


SuiYue-2308 (approved):

jonathangoeke (commented):
- does this only apply to junction correction? Not in other places in the code?
- are the results identical?

cying111:
The only place that uses xgboost function is here. It is internally called multiple times for splicing junction prediction and read class prediction.

It should be identical, as it's just a rename. But I am currently checking on a big data to see if results identical.

cying111:
Update: checked before and after the changes, se results are identical on a big sample.

jonathangoeke (approved):
Thanks for the change! Looks good. Can you include a note for the release?

PR #525, opened 2026-01-15
Fix xgboost parallel processing issue

cying111:
Fix the parallel processing issue with xgboost 
Now rurun bambu with the same R session with multiple cores will not cause hanging issue any more


jonathangoeke (approved):

SuiYue-2308 (approved):
I have tested it on a big data and the test data, the results are identical!

PR #526, opened 2026-01-15
Fix xgboost parallel issue for v4

cying111:
Fix the parallel processing issue with xgboost
Now rurun bambu with the same R session with multiple cores will not cause hanging issue any more


jonathangoeke (approved):

SuiYue-2308 (approved):
Have tested on large and small data. Results are identical!

PR #548, opened 2026-03-20
change priority in CB & UMI name extraction from bam file

lingminhao:


ch99l (changes requested):
Code is still incorrect. Issue likely relates to the generateColData function

PR #558, opened 2026-04-15
Summarise by exon

### LINKED ISSUE #563: summarise by exon count

ch99l:
## Related Issue
Fixes # (issue number)

## Type of Change
- [ ] **Bug fix** (non-breaking change which fixes an issue)
- [X] **New feature** (please specify if the change breaks existing functionality) 
    - [ ] **Non breaking change** (the feature doesn't change existing functionality)
    - [ ] **Breaking change**  (the feature that would cause existing functionality to not work as expected)
- [ ] **Documentation update**
- [ ] **Performance optimization**

## Description
Add summariseByExon() function (function is found in newly created summariseByExon.R script)                                                                                                                
                                                                                                                                                
Adds a new exported utility function summariseByExon() that aggregates transcript-level expression from a bambu SummarizedExperiment down to unique exon-level counts.                                                                                                                       
 
How it works:                                                                                                                                   
- Unlists the transcript GRangesList to get individual exon-transcript pairs                                                                  
- Defines unique exons by coordinate identity (seqnames:start:end:strand)                                                                       
- Builds a sparse binary membership matrix (unique exons × transcripts) and multiplies it against the transcript count matrix to efficiently sum counts across all transcripts sharing each exon                                                                                                
- Returns a RangedSummarizedExperiment with one row per unique exonic locus, preserving colData from the input and carrying GENEID (comma-separated for exons shared across genes) as row metadata                                                                                 
                                                                                                                                                
Returns: A RangedSummarizedExperiment where counts are for each unique exon across all overlapping transcripts.

## Implementation Details
Code change does not affect main bambu code as it is implemented as a standalone function in summariseByExon.R

## Impact of Changes
Briefly describe the impact of the implemented changes.
Generates a new SummarisedExperiment object containing transcript counts at exon level. Does not affect existing bambu codebase

## Checklist
- [X] My code follows the style guidelines of this project.
- [X] I have performed a self-review of my own code.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation (vignettes, man pages).
Note: Documentation will be updated later 
- [X] I have tested the code on a full dataset, and any differences have been described in the Impact of Changes section.


cying111:
I just asked claude code to review the PR and test the PR, so it created unit tests for this which I have already committed to the summarise-by-exon branch. 

And here is a summary of the issues identified by claude code, my summary is maybe can ignore the MInor issues, but the blocking issues and worth addressing issues can fix.

Prompt I used:
help me review this PR: https://github.com/GoekeLab/bambu/pull/558, document your footprint inside the folder, and test with small testdata

### Blocking / should fix before merge

1. **NAMESPACE and man page not regenerated.** The `@export` tag was added but
   `NAMESPACE` has no `export(summariseByExon)` and `man/` has no
   `summariseByExon.Rd`. Run `devtools::document()` / `roxygen2::roxygenise()`
   and commit the results — otherwise the function isn't actually exported
   from the installed package.
2. **No unit test.** `tests/testthat/` has a test file per exported function
   (e.g. `test_transcriptToGeneExpression.R`) but none for this one. A minimal
   test using the same fixture SE would match project conventions and catch
   regressions on the aggregation identity above.
3. **`assays = list(counts = …)`** in the returned SE — the rest of the
   codebase uses `S4Vectors::SimpleList(...)` (see
   `R/transcriptToGeneExpression.R:47`). Works (SE coerces) but inconsistent.
   Copilot flagged the same.

### Worth addressing

4. **No input validation.** The function assumes `rowData(se)$GENEID` exists
   and that `assays(se)$counts` is present and numeric. Passing anything else
   fails with confusing downstream errors. A short
   `stopifnot(is(se, "RangedSummarizedExperiment"), "counts" %in% assayNames(se), "GENEID" %in% colnames(rowData(se)))`
   would be enough.
5. **Ignores `metadata(se)$incompatibleCounts` and `$nonuniqueCounts`.**
   `transcriptToGeneExpression` folds these in before aggregating. Whether
   that's desired here is a design call, but at minimum the docstring should
   say the exon-level counts do *not* include incompatible/nonunique reads,
   so users aren't surprised that `sum(exonCounts) != sum(assayed reads)`.
6. **No `CPM` assay.** `transcriptToGeneExpression` returns counts + CPM; this
   returns counts only. Fine, but worth being explicit in the doc.
7. **Roxygen `@examples` missing.** Other exported functions carry a runnable
   example that loads the fixture SE — easy to add here and gives R CMD check
   coverage of the example path.
8. **`@import data.table`** — `transcriptToGeneExpression` does the same, so
   this is just matching existing style, but Bioconductor review usually
   prefers `@importFrom data.table data.table :=` to narrow the import.

### Minor

9. Exon rownames are `chr:start:end:strand` strings — readable but long. Fine
   as-is; document the format.
10. If a transcript's `rowRanges` has no names, `unlist(..., use.names = TRUE)`
    produces `NA` names, which would then cascade into `exonDt$TXNAME`. Not a
    real concern for bambu output (txs are always named) but trivially
    defensible via the validation step above.
11. PR description says "changes remain isolated … do not modify core Bambu
    code." True for the code, but NAMESPACE/man files *do* need to change.

jonathangoeke:
### Code review

Found 8 issues (scored 0-100 for confidence): (JG:threshold is 80, all below threshold)

1. **NAMESPACE not regenerated -- function is not actually exported** (score: 75). `@export` tag is present but `devtools::document()` / `roxygen2::roxygenise()` was not run. No `export(summariseByExon)` in NAMESPACE and no `man/summariseByExon.Rd`. The function will not be accessible after `library(bambu)`.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L13-L15

2. **Genomic sort removed during refactoring -- regression** (score: 75). Commit b4277ca had `result <- result[order(seqnames, start, end)]`. Commit dbe2af9 refactored to SummarizedExperiment and silently dropped the sort. Output is now in data.table first-occurrence order, not genomic order. Bioconductor convention expects sorted GRanges in a RangedSummarizedExperiment.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L65-L81

3. **Only `counts` assay propagated; `CPM`, `fullLengthCounts`, `uniqueCounts` silently dropped** (score: 50). `transcriptToGeneExpression` returns both `counts` and `CPM`. This function returns only `counts` without warning. May be intentional but is undocumented divergence from the analogous function.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L62-L64

4. **`list()` instead of `SimpleList()` for assays** (score: 50). Every other SE constructor in the codebase uses `S4Vectors::SimpleList`. Even this function's own test uses `SimpleList`. Functionally equivalent due to auto-coercion, but inconsistent with codebase convention.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L76-L78

5. **`@return` claims `RangedSummarizedExperiment` but class is not imported** (score: 50). The `@importFrom` on line 11 imports only `SummarizedExperiment` (the constructor). The analogous function avoids this by saying `@return A SummarizedExperiment object`. Minor documentation pedantry.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L3-L11

6. **No `@examples` block for exported function** (score: 25). Every other exported function in the package has `@examples`. BiocCheck will warn. The test file already contains the exact idiom needed.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L1-L14

7. **`incompatibleCounts` / `nonuniqueCounts` not included** (score: 25). `transcriptToGeneExpression` adds these before aggregating. However, since incompatible reads cannot be attributed to specific exons, omitting them is arguably correct.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L62-L64

8. **`colData` rownames not realigned to assay column names** (score: 0, false positive). Matrix multiplication preserves column names, so `colData(se)` rownames already match.

https://github.com/GoekeLab/bambu/blob/e6c55473021588cc40cb285539320311f506fe63/R/summariseByExon.R#L78-L80

jonathangoeke (commented):
**specific comments:**
 -names are not very informative. do we need names? this should ideally be official exonIds, which we might not track in bambu? align with other summarisefunctions
 -mcols: need to have a map of exon to transcripts. 
 -mcols: add if novel exon or annotated exon (only present in novel transcripts vs at least one annotated transcript)
 -provide option which assays to summarise (by default all?)

  **general comments/combine this PR with other updates later?**
 - do not add extra file. instead combined with existing file. rename file to summariseExpression, and summariseExpression_utlity?
 - make sure that code is consistently implemented, use same style reuse code if possible (e.g same as transcriptToGeneexpression)
 - add example in documentation/ user facing functions should be well documented
 - add documentation, or create draft issue that this needs to be documented later
 - one function to summariseExpression('exons/transcript/start/end')? or individual functions? they should use same structure to avoid uplicated code. e.g define summary-map, summarise assays, summarise ranges, summarise mcol, ...? if possible
 - use dplyr to be consistent with major code in bambu
 
 
 **unit test:** 
 -document unit test lines, what is tested? why is this relevant? if the unit test function needs to be clear otherwise not clear if fails and difficult to maintain
 - unit test that just reimplement the code are not so useful. It might be helpful to define a manual examples of 2 genes with 1 trancript and with 3 transcripts and multpile exons, and transcripts counts, and define expected output e.g. 5 rows, 3 read counts
 -unit test should cover multiple samples (min 2)
 -current unit test object has zero count is that meaningful?

R/summariseByExon.R
```diff
@@ -0,0 +1,81 @@
...
+#' @importFrom IRanges IRanges
+#' @export
+summariseByExon <- function(se) {
+    # Unlist GRangesList: one row per exon-transcript combination
+    exonRanges <- unlist(rowRanges(se), use.names = TRUE)
+
+    # Build a data.table of exon-transcript pairs
+    txNames <- rownames(se)
+    exonDt <- data.table(
+        TXNAME   = names(exonRanges),
```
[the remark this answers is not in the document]
cying111: i think original code is good, names(exonRanges) is simpler

R/summariseExpression.R
```diff
@@ -0,0 +1,91 @@
...
+    # Build data.frame of exon-transcript pairs
+    exonDf <- data.frame(
+        TXNAME   = names(exonRanges),
+        seqnames = as.character(seqnames(exonRanges)),
+        start    = start(exonRanges),
+        end      = end(exonRanges),
+        strand   = as.character(strand(exonRanges)),
+        stringsAsFactors = FALSE
+    ) |>
+        mutate(exon_id = paste(seqnames, start, end, strand, sep = ":"))
```
jonathangoeke: %>% group_by(...) + create index

R/summariseExpression.R
```diff
@@ -0,0 +1,91 @@
...
+
+    # Attach GENEID and txClassDescription from rowData
+    geneDf <- as.data.frame(rowData(se))[, c("GENEID", "txClassDescription"), drop = FALSE]
```
jonathangoeke: novelTranscripts should be there

R/summariseExpression.R
```diff
@@ -0,0 +1,91 @@
...
+            txNames  = paste(sort(unique(TXNAME)),  collapse = ","),
+            .groups  = "drop"
+        )
+
+    # exonClass: "annotated" if the exon coordinates (seqnames, start, end, strand)
+    # exist in the reference annotation, "novel" otherwise.
+    annotatedCoords <- exonDf |>
+        filter(txClassDescription == "annotation") |>
+        distinct(seqnames, start, end, strand) |>
+        mutate(exonClass = "annotated")
```
jonathangoeke: use novelTranscript as part of summarise above

R/summariseExpression.R
cying111: isn't TXNAME already in rowData column?

R/summariseExpression.R
```diff
@@ -0,0 +1,91 @@
...
+    # Attach GENEID and txClassDescription from rowData
+    geneDf <- as.data.frame(rowData(se))[, c("GENEID", "txClassDescription"), drop = FALSE]
+    geneDf$TXNAME <- rownames(se)
+    exonDf <- left_join(exonDf, geneDf, by = "TXNAME")
+
+    # Collapse coordinates and GENEID/txNames per unique exon
+    exonMeta <- exonDf |>
+        group_by(exon_id) |>
+        summarise(
+            seqnames = seqnames[1],
```
cying111: in fact, if exon_id can be used, why not group_by exon_id, seqnames, start, end and strand? because they would just be shared for the same exon_id

R/summariseExpression.R
jonathangoeke: use all()

R/summariseExpression.R
```diff
@@ -0,0 +1,91 @@
...
+        distinct(seqnames, start, end, strand) |>
+        mutate(exonClass = "annotated")
+
+    exonMeta <- exonMeta |>
+        left_join(annotatedCoords, by = c("seqnames", "start", "end", "strand")) |>
+        mutate(exonClass = ifelse(is.na(exonClass), "novel", "annotated"))
+
+    # Sparse binary matrix: unique_exons x transcripts
+    uniqueExons <- exonMeta$exon_id
+    exonNames   <- paste0("EX", seq_along(uniqueExons))
```
jonathangoeke: generalise by groupID

R/summariseExpression.R
```diff
@@ -0,0 +1,91 @@
...
+        dimnames = list(exonNames, txNames)
+    )
+
+    # Build output GRanges
+    outRanges <- GRanges(
+        seqnames = exonMeta$seqnames,
+        ranges   = IRanges(start = exonMeta$start, end = exonMeta$end),
+        strand   = exonMeta$strand
+    )
+    names(outRanges) <- exonNames
```
jonathangoeke: groupNames

jonathangoeke (commented):
generalise, see comments

R/summariseExpression.R
```diff
@@ -0,0 +1,91 @@
...
+    mcols(outRanges)$GENEID  <- exonMeta$GENEID
+    mcols(outRanges)$txNames <- exonMeta$txNames
+    mcols(outRanges)$exonClass <- exonMeta$exonClass
+
+    .aggregateAssays(se, aggregationMat, outRanges)
```
cying111: where is this function loaded from? there is a same function from library(SingleCellExperiments), but I am not sure whether it does the same thing, probably not (gemini suggested code for this function)

Assuming 'se' is your SingleCellExperiment object
and 'cluster_id' is a column in colData(se)
library(SingleCellExperiment)
aggregated_se <- .aggregateAssays(se, 
                                  grouping = colData(se)$cluster_id, 
                                  fun = "mean")

trying to find it, but not able to find it anywhere

I think maybe ok to leave it, looks very different in terms of usage

PR #559, opened 2026-04-15
fix: adjust NDR threshold behavior. Previously, NDR = 1 would overrid…

### LINKED ISSUE #560: refine the NDR usage
This issue is related to refine the NDR usage. 
1. currently,  NDR = 1 might include the low confidence transcripts
1. check the difference between NDR = 0.999 and NDR = 1 
2. refine the NDR that filtering will only be applied to non low confidence and non subset transcript.  
3. include the low confidence confidence and subet transcript can be another 2 parameters.

SuiYue-2308:
fix: adjust NDR threshold behavior

## Type of Change
- [x] **Bug fix** (non-breaking change which fixes an issue)

## Description
Previously, NDR = 1 would override `remove.subset` and `min.sampleNumber`. Now, NDR = 0.999 and NDR = 1 produce identical results except for the NDR threshold itself.

## Implementation Details
Before, if the read class is a subset or low confidence transcript (can not pass the min.sampleNumber), it's NDR will be assigned to 1, which will be further reported when setting NDR = 1. Now, I set their NDR to NA, and add another check step before report novel transcript (!is.na(NDR)) to make sure the subset and low confidence are not included.

## Impact of Changes
Briefly describe the impact of the implemented changes. For example: 
- Results are identical to the previous version when NDR = 0.999.
- Now "NDR = 1" will not report subset and low confidence transcripts

## Checklist
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have tested the code on a full dataset, and any differences have been described in the Impact of Changes section.

### DIFF
--- a/R/bambu-extendAnnotations-utilityExtend.R
+++ b/R/bambu-extendAnnotations-utilityExtend.R
@@ -827,7 +827,6 @@ addGeneIdsToReadClassTable <- function(readClassTable, distTable,
 #' @description This function train a model for use on other data
 #' @param extendedAnnotations A GRangesList object produced from bambu(quant = FALSE) or rowRanges(se)
 #' @param NDR The maximum NDR for novel transcripts to be in extendedAnnotations (0-1). If not provided a recommended NDR is calculated.
-#' @param includeRef A boolean which if TRUE will also filter out reference annotations based on their NDR
 #' @param prefix A string which determines which transcripts are considered novel by bambu and will be filtered (by default = 'Bambu')
 #' @param baselineFDR a value between 0-1. Bambu uses this FDR on the trained model to recommend an equivilent NDR threshold to be used for the sample. By default, a baseline FDR of 0.1 is used. This does not impact the analysis if an NDR is set.
 #' @param defaultModels a bambu trained model object that bambu will use when fitReadClassModel==FALSE or the data is not suitable for training, defaults to the pretrained model in the bambu package
@@ -835,7 +834,7 @@ addGeneIdsToReadClassTable <- function(readClassTable, distTable,
 #' @details 
 #' @return extendedAnnotations with a new NDR threshold
 #' @export
-setNDR <- function(extendedAnnotations, NDR = NULL, includeRef = FALSE, prefix = 'Bambu', baselineFDR = 0.1, defaultModels2 = defaultModels){
+setNDR <- function(extendedAnnotations, NDR = NULL, prefix = 'Bambu', baselineFDR = 0.1, defaultModels2 = defaultModels){
     #Check to see if the annotations/gtf are dervived from Bambu
     if(is.null(mcols(extendedAnnotations)$NDR)){
         warning("Annotations were not extended by Bambu (or the wrong prefix was provided). NDR can not be set")
@@ -852,17 +851,10 @@ setNDR <- function(extendedAnnotations, NDR = NULL, includeRef = FALSE, prefix =
         message("Recommending a novel discovery rate (NDR) of: ", NDR)
     }
 
-    #If reference annotations should be filtered too (note that reference annotations with no read support arn't filtered)
-    if(includeRef){
-        toRemove <- (!is.na(mcols(extendedAnnotations)$NDR) & mcols(extendedAnnotations)$NDR > NDR)
-        toAdd <- !is.na(mcols(metadata(extendedAnnotations)$lowConfidenceTranscripts)$NDR) & 
-            mcols(metadata(extendedAnnotations)$lowConfidenceTranscripts)$NDR <= NDR  
-    } else {
-        toRemove <- (mcols(extendedAnnotations)$NDR > NDR & 
-            grepl(prefix, mcols(extendedAnnotations)$TXNAME))
-        toAdd <- (mcols(metadata(extendedAnnotations)$lowConfidenceTranscripts)$NDR <= NDR & 
-            grepl(prefix, mcols(metadata(extendedAnnotations)$lowConfidenceTranscripts)$TXNAME))     
-    }
+    toRemove <- (mcols(extendedAnnotations)$NDR > NDR &
+        grepl(prefix, mcols(extendedAnnotations)$TXNAME))
+    toAdd <- (mcols(metadata(extendedAnnotations)$lowConfidenceTranscripts)$NDR <= NDR &
+        grepl(prefix, mcols(metadata(extendedAnnotations)$lowConfidenceTranscripts)$TXNAME))
   
   temp <- c(metadata(extendedAnnotations)$lowConfidenceTranscripts[!toAdd], extendedAnnotations[toRemove])
   extendedAnnotations <- c(extendedAnnotations[!toRemove], metadata(extendedAnnotations)$lowConfidenceTranscripts[toAdd])
--- a/R/bambu-extendAnnotations-utilityExtend.R
+++ b/R/bambu-extendAnnotations-utilityExtend.R
@@ -112,7 +112,7 @@ filterTranscriptsByAnnotation <- function(rowDataCombined, annotationGrangesList
   } else if(is.null(NDR)) {
           NDR <- 0.5
   }
-  filterSet <- (rowDataCombined$NDR <= NDR | rowDataCombined$readClassType == "equal:compatible")
+  filterSet <- ((!is.na(rowDataCombined$NDR) & rowDataCombined$NDR <= NDR) | rowDataCombined$readClassType == "equal:compatible")
   lowConfidenceTranscripts <- combindRowDataWithRanges(
         rowDataCombined[!filterSet,], 
         exonRangesCombined[!filterSet])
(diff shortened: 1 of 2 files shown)


cying111:
Adding in my PR review comment here (also with assistance of claude code), my summary is issue 1, 3 worth noting, issue 4 is a comment adding, maybe good to have, but also fine to not have 

Prompt used:
now help me review this PR: https://github.com/GoekeLab/bambu/pull/559
### Worth addressing before merge

1. **`bambu.R:164` sets `NDR <- 1` for `mode == "fusion"`.** This fix
   therefore changes fusion-mode behaviour: fusion mode now respects
   `remove.subsetTx` and `min.sampleNumber`, where previously it didn't.
   That may be correct (and may be exactly what #560 wants) but it's not
   mentioned in the PR description. Either:
   - confirm it's intended and call it out in the PR body / NEWS, or
   - switch fusion mode to a sentinel that preserves the old permissive
     behaviour (e.g. pass `NDR = Inf` plus a separate `remove.subsetTx =
     FALSE` in fusion mode).

2. **No test.** The PR's central claim — "NDR = 0.999 and NDR = 1 produce
   identical results except for the NDR threshold itself" — is easy to
   encode as a testthat expectation against the existing fixture
   (`seIsoReCombined_SGNex_..._chr9_1_1000000.rds` + the annotation GR).
   The table above is effectively that test; porting it would guard against
   a silent regression.

3. **`test_setNDR.R` line 68–82 (pre-existing).** The block
   `test_that("setNDR works when no lowConfidenceTranscripts are present", ...)`
   computes `extendedAnnotations` via `isore.extendAnnotations(..., NDR = 1, ...)`
   but then asserts on the pre-saved `extendedAnnotationsExpected` fixture
   — the freshly computed `extendedAnnotations` is never used. Not a
   blocker for this PR, but this was the test most likely to have caught
   the pre-fix bug, and it silently passed because of that oversight. A
   follow-up to rewire that test against the computed object is worth a
   separate commit.

### Minor

4. The new predicate `!is.na(x) & x <= NDR` is a clean way to kill off NAs,
   but it's worth a brief code comment that NA NDR = "deliberately filtered"
   (subset or scoreless) so a future reader doesn't "simplify" it back.

jonathangoeke:
### Code review

Top 3 issues by confidence score (lower threshold than usual):

1. **(score: 75) `setNDR()` `includeRef=FALSE` branch lacks `!is.na()` guards.** With this PR, transcripts in `lowConfidenceTranscripts` can have `NDR = NA`. The `includeRef=TRUE` branch guards against NA, but the default `includeRef=FALSE` branch does not — `NA <= NDR` propagates NA into `toAdd`/`toRemove`, then into logical subsetting of the GRangesList, which can produce corrupted/NA entries.

https://github.com/GoekeLab/bambu/blob/25aa6794805f82d64206245bd7bb76e4273c8772/R/bambu-extendAnnotations-utilityExtend.R#L855-L866

2. **(score: 50) Fusion mode side effect.** `bambu.R` sets `NDR <- 1` for fusion mode. Combined with the new filter at `utilityExtend.R:115` (which now drops `NDR == NA` rows), fusion-mode output may differ from prior behavior in ways that aren't covered by tests.

https://github.com/GoekeLab/bambu/blob/25aa6794805f82d64206245bd7bb76e4273c8772/R/bambu.R#L161-L167

3. **(score: 25) Semantic change to prior intent.** Commit [`0b47cb8`](https://github.com/GoekeLab/bambu/commit/0b47cb8) introduced `NDR <- 1` for subset/low-confidence transcripts with the stated purpose "Report all possible transcripts if NDR = 1". This PR reverses that intent. Likely intentional given the PR description, but worth calling out explicitly so downstream users relying on `NDR = 1` to recover all candidates know the behavior has changed.

https://github.com/GoekeLab/bambu/blob/25aa6794805f82d64206245bd7bb76e4273c8772/R/bambu-extendAnnotations-utilityExtend.R#L223-L229

Only #1 would have cleared the default score threshold (80); #2 and #3 are included at the user's request for a lower-threshold review.

jonathangoeke (commented):
Did you confirm that you can reproduce results with NDR=1 (previous version) and NDR=1 (including subset and low confidence read classes)? Which parameters need to be selected to reproduce this behaviour?

SuiYue-2308:
> Did you confirm that you can reproduce results with NDR=1 (previous version) and NDR=1 (including subset and low confidence read classes)? Which parameters need to be selected to reproduce this behaviour?

Hi @jonathangoeke, Yes, we can get the exact same result with previous NDR = 1 by setting `opt.discovery = list(remove.subsetTx = FALSE, min.readCount = 1, min.sampleNumber = 0)` to include the subset and low confidence.

R/bambu-extendAnnotations-utilityExtend.R
```diff
@@ -224,7 +224,7 @@ calculateNDROnTranscripts <- function(combinedTranscripts, useTxScore = FALSE){
     } else {
         combinedTranscripts$NDR <- calculateNDR(combinedTranscripts$maxTxScore, equal)
     }
-    combinedTranscripts$NDR[combinedTranscripts$maxTxScore==-1] <- 1
+    combinedTranscripts$NDR[combinedTranscripts$maxTxScore==-1] <- NA
```
cying111, line 227: I have a similar comment as copilot, if set NDR equal to NA, then following commands related to NDR need to be NA aware. Need to check on this, and make sure all following commands pass.

cying111: Other than this, different scenarios have been tested, and all works as expected:
testing for isore.extendAnnotations reported set of transcripts are expectedly corrected to have subsetTranscripts kept even if rm.subsetTx = TRUE and all filtering pre-set thresholds
testing for whole bambu shows that quantification results impacted, cause reporting set of transcripts changed at last 

Also one thing is that this fix does not only fix the override of return.subsetTx and min.SampleNumber, and also other base filtering, maybe can consider to make this more explicit

SuiYue-2308:
Hi @cying111 @jonathangoeke 

Following the comments, I updated this pull request. 
1. **Fusion mode**: previously, the fusion mode will set NDR = 1, which will include everything. Now I add the following setting to the fusion mode to return the same result with fusion mode. 
```
opt.discovery$remove.subsetTx <- FALSE
opt.discovery$min.readCount <- 1
opt.discovery$min.sampleNumber <- 0
```
I have tested the fusion mode on a real data set, the rowRanges and the assays are identical, the only thing is that this branch not keep the subset in the metadata(rowRanges(se)). If we want to keep subset need to fix the fusion mode in future. 

2. **includeRef**
I have removed this parameter since we never use this. 

3. the third part from @jonathangoeke   generated by code agent is not that clear to me.

cying111:
Hi @SuiYue-2308 , I think you are right, min.readCount is needed only for the change, but the se object changed in the end, 
I think NDR, maxTxScore, maxTxScore.noFit changes are expected because change from 1 to NA, would cause these differences 
The rowRanges(se_v4)$metadata vs rowRanges(se_pr)$metadata are details below as well, which I think are expected as well.
 
Error: Expected `se_v4` to equal `se_pr`.
Differences:
Attributes: < Component “rowRanges”: Attributes: < Component “elementMetadata”: Attributes: < Component “listData”: Component “NDR”: Mean relative difference: 0.1815954 > > >
Attributes: < Component “rowRanges”: Attributes: < Component “elementMetadata”: Attributes: < Component “listData”: Component “maxTxScore”: Mean relative difference: 1.181595 > > >
Attributes: < Component “rowRanges”: Attributes: < Component “elementMetadata”: Attributes: < Component “listData”: Component “maxTxScore.noFit”: Mean relative difference: 1.181595 > > >
Attributes: < Component “rowRanges”: Attributes: < Component “metadata”: Names: 1 string mismatch > >
Attributes: < Component “rowRanges”: Attributes: < Component “metadata”: Length mismatch: comparison on first 2 components > >
Attributes: < Component “rowRanges”: Attributes: < Component “metadata”: Component 2: Lengths: 17, 0

```
> rowRanges(se_v4)@metadata
$NDRthreshold
[1] 1

$subsetTranscripts
GRangesList object of length 17:
$BambuTx1
GRanges object with 3 ranges and 2 metadata columns:
      seqnames      ranges strand | exon_rank exon_endRank
         <Rle>   <IRanges>  <Rle> | <integer>    <integer>
  [1]        9 14513-16876      - |         3            1
  [2]        9 16969-17166      - |         2            2
  [3]        9 17344-17410      - |         1            3
  -------
  seqinfo: 1 sequence from an unspecified genome; no seqlengths

$BambuTx3
GRanges object with 3 ranges and 2 metadata columns:
      seqnames      ranges strand | exon_rank exon_endRank
         <Rle>   <IRanges>  <Rle> | <integer>    <integer>
  [1]        9 14517-14940      - |         3            1
  [2]        9 15081-15149      - |         2            2
  [3]        9 15909-16064      - |         1            3
  -------
  seqinfo: 1 sequence from an unspecified genome; no seqlengths

$BambuTx5
GRanges object with 2 ranges and 2 metadata columns:
      seqnames      ranges strand | exon_rank exon_endRank
         <Rle>   <IRanges>  <Rle> | <integer>    <integer>
  [1]        9 14521-14940      - |         2            1
  [2]        9 15081-15750      - |         1            2
  -------
  seqinfo: 1 sequence from an unspecified genome; no seqlengths

...
<14 more elements>

$lowConfidenceTranscripts
GRangesList object of length 0:
<0 elements>

> rowRanges(se_pr)@metadata
$NDRthreshold
[1] 1

$lowConfidenceTranscripts
GRangesList object of length 0:
<0 elements>
```

cying111 (approved):

jonathangoeke (approved):

SuiYue-2308:
Thanks for the code review!

I will merge this pull request now, and we will need to look into the metadata to decide whether we still want to keep the subset and low confidence transcript.

PR #567, opened 2026-04-16
 Added functionality for writeBambuOutput() to write to gff3

### LINKED ISSUE #564: add functionality to write to GFF file
currently: writeToGTF
add functionality to write to GFF3 format.
define function "writeAnnotations" that uses either write to GTF or write to GFF3

Bambu typically only writes transcript & exon.

1. Load a bambu-results.rds file (reuse from previous minimal SG-NEx Bambu run on chr22)
2. Call writeBambuOutput and see how it calls writeToGTF
3. Define writeToGFF3 helper function based on the above
4. Define writeAnnotations
5. adjust writeBambuOutput to take writeAnnotations

**Readings**
[GTF (Ensembl)](https://www.ensembl.org/info/website/upload/gff.html?)
[GTF (UCSC)](https://genome.ucsc.edu/FAQ/FAQformat.html#format4)

[GFF3 (Ensembl)](https://www.ensembl.org/info/website/upload/gff3.html)

[A comprehensive comparison (AGAT)](https://agat.readthedocs.io/en/latest/index.html)

[On converting GTF to GFF3:](https://gmod.org/wiki/GFF2#Converting_GFF2_to_GFF3)

**General observations**
GTF is equivalent to GFF2.
GFF3 can have file extension gff or gff3
Both are 9-column tab-delimited tabular files

GTF differs from GFF3 mainly in the 9th column (attribute) and the syntax of the key:value pairs.

GTF:
```
gene_id "GENE1"; transcript_id "TX1", gene_name "BRCA1";
gene_id "GENE1"; transcript_id "TX1", exon_number "1";
```

GFF3:
```
ID=TX1;Parent=GENE1;Name=BRCA1;
ID=exon1;Parent=TX1;exon_number=1;
```
Documentation on what these mean can be found [here](https://gmod.org/wiki/GFF3#GFF3_Format) 

Also, GTF attributes have a trailing `;` whereas GFF3 does not.

Other differences include
- GFF3 header
- GTF column 3 (feature) is constrained to lesser terms than GFF3

hafiz-ismail:
## Related Issue
Fixes # (issue number)

## Type of Change
- [ ] **Bug fix** (non-breaking change which fixes an issue)
- [x] **New feature** (please specify if the change breaks existing functionality) 
    - [ ] **Non breaking change** (the feature doesn't change existing functionality)
    - [x] **Breaking change**  (the feature that would cause existing functionality to not work as expected)
- [x] **Documentation update**
- [ ] **Performance optimization**

## Description
Implemented functionality for writeBambuOutput to export gff3 files from SummarizedExperiment objects. Currently Bambu can export the more stringent gtf file format, but there is no functionality for gff3. An overview of the differences of the 2 file formats:

GTF differs from GFF3 mainly in the 9th column (attribute) and the syntax of the key:value pairs.

GTF:
```
gene_id "GENE1"; transcript_id "TX1"; gene_name "BRCA1";
gene_id "GENE1"; transcript_id "TX1"; exon_number "1";
```
[GTF documentation (Ensembl)](https://asia.ensembl.org/info/website/upload/gff.html?)

GFF3:
```
ID=TX1;Parent=GENE1;Name=BRCA1
ID=exon1;Parent=TX1;exon_number=1
```

[GFF3 documentation (Ensembl)](https://asia.ensembl.org/info/website/upload/gff3.html)

Also, GTF attributes have a trailing `;` whereas GFF3 does not.

Other differences include
- GFF3 header
- GTF column 3 (feature) is constrained to lesser terms than GFF3

## Implementation Details
All code changes were implemented in readWrite.R.
- Implemented two new helper functions: writeToGFF3 and writeAnnotationsToGFF3, following convention from already existing writeToGTF and writeAnnotationsToGTF
- Implemented new parameter `outputAnnFormat` in writeBambuOutput: takes either `"gtf"` (default) or `"gff3"` as arguments
- renamed `transcript_gtffn` to `transcript_annfn`
- Updated documentation in README and comment markers parsed by roxygen2

## Impact of Changes
- All previous functionality is preserved, writeBambuOutput still exports to .gtf by default unless specified otherwise. It was found that the current writeBambuOutput in devel_pre_v4 is not fully functional - therefore a minimal function is provided to test results as below:

```
writeBambuOutput <- function(se, path, prefix = "", outputExtendedAnno = TRUE, 
                             outputAll = TRUE, outputBambuModels = TRUE, outputNovelOnly = TRUE, seperateSamples = FALSE,
                             outputAnnFormat = "gtf") {
  
  if (missing(se) | missing(path)) {
    stop("Both summarizedExperiment object from bambu and
            the path for the output files are required.")
  } else if (!outputAnnFormat %in% c("gtf", "gff3")) {
    stop("Please specify a valid annotation outputAnnFormat: 'gtf' or 'gff3', or leave blank to output to GTF by default.")
  } else {
    outdir <- paste0(path, "/")
    if (!dir.exists(outdir))
      dir.create(outdir, recursive = TRUE)
    
    transcript_grList <- rowRanges(se)
    prefix <- ifelse(prefix != "", paste0(prefix, "_"), "")
    transcript_annfn <- paste(outdir, prefix, sep = "")
    
    if (outputAnnFormat == "gtf") {
      gtf <- writeAnnotationsToGTF(annotation = transcript_grList,
                                   file = transcript_annfn, outputExtendedAnno = outputExtendedAnno, 
                                   outputAll = outputAll, outputBambuModels = outputBambuModels, outputNovelOnly = outputNovelOnly)
    }
    
    if (outputAnnFormat == "gff3") {
      gff3 <- writeAnnotationsToGFF3(annotation = transcript_grList,
                                     file = transcript_annfn, outputExtendedAnno = outputExtendedAnno, 
                                     outputAll = outputAll, outputBambuModels = outputBambuModels, outputNovelOnly = outputNovelOnly)
    }
  }
}
```

In addition, this output has not been updated for compatibility with importBambuResults - providing functionality for this will be associated with future implementations for the prepareAnnotations function

## Checklist
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation (vignettes, man pages).
- [x] I have tested the code on a full dataset, and any differences have been described in the Impact of Changes section.

### DIFF
--- a/R/readWrite.R
+++ b/R/readWrite.R
@@ -4,10 +4,12 @@
 #' @param path the destination of the output files 
 #' (gtf, transcript counts, and gene counts)
 #' @param prefix the prefix of the output files
-#' @details The function will write the output from Bambu to files. The 
-#' annotations will be written to a .gtf file, transcript counts (total counts, 
+#' @param outputAnnFormat the file format in which to output annotations, must
+#' be one of \code{"gtf"} or \code{"gff3"}. \code{"gtf"} is specified by default.
+#' @details The function will write the output from Bambu to files. The
+#' annotations will be written to a .gtf or .gff3 file, transcript counts (total counts, 
 #' CPM, full-length counts, partial-length counts, and unique counts) and gene counts
-#' will be written to .txt files.  
+#' will be written to .txt files.
 #' @export
 #' @examples
 #' se <- readRDS(system.file("extdata",
@@ -17,23 +19,36 @@
 #' path <- tempdir()
 #' writeBambuOutput(se, path)
 writeBambuOutput <- function(se, path, prefix = "", outputExtendedAnno = TRUE, 
-                             outputAll = TRUE, outputBambuModels = TRUE, outputNovelOnly = TRUE, seperateSamples = FALSE) {
-    if (missing(se) | missing(path)) {
-        stop("Both summarizedExperiment object from bambu and
+                             outputAll = TRUE, outputBambuModels = TRUE, outputNovelOnly = TRUE, seperateSamples = FALSE,
+                             outputAnnFormat = "gtf") {
+  
+  if (missing(se) | missing(path)) {
+    stop("Both summarizedExperiment object from bambu and
             the path for the output files are required.")
-    } else {
-        outdir <- paste0(path, "/")
-        if (!dir.exists(outdir))
-            dir.create(outdir, recursive = TRUE)
-
-        transcript_grList <- rowRanges(se)
-        prefix <- ifelse(prefix != "", paste0(prefix, "_"), "")
-        transcript_gtffn <- paste(outdir, prefix, sep = "")
-        gtf <- writeAnnotationsToGTF(annotation = transcript_grList,
-            file = transcript_gtffn, outputExtendedAnno = outputExtendedAnno, 
-            outputAll = outputAll, outputBambuModels = outputBambuModels, outputNovelOnly = outputNovelOnly)
+  } else if (!outputAnnFormat %in% c("gtf", "gff3")) {
+    stop("Please specify a valid annotation outputAnnFormat: 'gtf' or 'gff3', or leave blank to output to GTF by default.")
+  } else {
+    outdir <- paste0(path, "/")
+    if (!dir.exists(outdir))
+      dir.create(outdir, recursive = TRUE)
+    
+    transcript_grList <- rowRanges(se)
+    prefix <- ifelse(prefix != "", paste0(prefix, "_"), "")
+    transcript_annfn <- paste(outdir, prefix, sep = "")
+    
+    if (outputAnnFormat == "gtf") {
+      gtf <- writeAnnotationsToGTF(annotation = transcript_grList,
+                                   file = transcript_annfn, outputExtendedAnno = outputExtendedAnno, 
+                                   outputAll = outputAll, outputBambuModels = outputBambuModels, outputNovelOnly = outputNovelOnly)
+    }
+    
+    if (outputAnnFormat == "gff3") {
+      gff3 <- writeAnnotationsToGFF3(annotation = transcript_grList,
+                                     file = transcript_annfn, outputExtendedAnno = outputExtendedAnno, 
+                                     outputAll = outputAll, outputBambuModels = outputBambuModels, outputNovelOnly = outputNovelOnly)
+    }
         
-        utils::write.table(colData(se), file = paste0(transcript_gtffn, "sampleData.tsv"), 
+        utils::write.table(colData(se), file = paste0(transcript_annfn, "sampleData.tsv"), 
             sep = "\t", quote = FALSE, row.names = FALSE, col.names = TRUE)
         for(d in names(assays(se))){
             writeCountsOutput(se, varname=d,
@@ -43,7 +58,7 @@ writeBambuOutput <- function(se, path, prefix = "", outputExtendedAnno = TRUE,
         #write incompatible counts
         if(!is.null(metadata(se)$incompatibleCounts)){
             estimates <- metadata(se)$incompatibleCounts
-            estimatesfn <- paste(transcript_gtffn, "incompatibleCounts.mtx", sep = "")
+            estimatesfn <- paste(transcript_annfn, "incompatibleCounts.mtx", sep = "")
                 Matrix::writeMM(estimates, estimatesfn)
         }
         seGene <- transcriptToGeneExpression(se)
@@ -51,9 +66,9 @@ writeBambuOutput <- function(se, path, prefix = "", outputExtendedAnno = TRUE,
         #utils::write.table(paste0(colnames(se), "-1"), file = paste0(outdir, "barcodes.tsv"), quote = FALSE, row.names = FALSE, col.names = FALSE)
         #R.utils::gzip(paste0(outdir, "barcodes.tsv"))
         txANDGenes <- data.table(as.data.frame(rowData(se))[,c("TXNAME","GENEID")])
-        utils::write.table(txANDGenes, file = paste0(transcript_gtffn, "txANDgenes.tsv"), 
+        utils::write.table(txANDGenes, file = paste0(transcript_annfn, "txANDgenes.tsv"), 
                            sep = "\t", quote = FALSE, row.names = FALSE, col.names = FALSE)
-        utils::write.table(names(seGene), file = paste0(transcript_gtffn, "genes.tsv"), 
+        utils::write.table(names(seGene), file = paste0(transcript_annfn, "genes.tsv"), 
                         sep = "\t", quote = FALSE, row.names = FALSE, col.names = FALSE)
         
         #R.utils::gzip(paste0(outdir, "txANDgenes.tsv"))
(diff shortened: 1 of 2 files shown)


jonathangoeke:
### Code review

Found 3 issues:

1. `writeToGFF3()` dfTx block accesses `mcols(annotation)$txScore` and `mcols(annotation)$txScore.noFit`, but these columns do not exist on the annotation GRangesList. The exon block in the same function correctly uses `mcols(annotation)$maxTxScore` and `mcols(annotation)$maxTxScore.noFit` (lines 290-291). This will produce `NULL` values in transcript-level GFF3 attributes. (bug: wrong column name, copy-pasted from the same pre-existing bug in `writeToGTF` at lines 199-200)

https://github.com/GoekeLab/bambu/blob/67f184d736f5b8dca55b5f77b2b5c94a1c226580/R/readWrite.R#L338-L341

2. `writeToGFF3()` writes `mRNA` records with `Parent=GENE_ID` and `exon` records with `Parent=TX_ID`, but no gene-level records are written. The GFF3 specification requires a three-level hierarchy (gene -> mRNA -> exon) where every `Parent` attribute references an `ID` that exists in the file. Without gene records, every `mRNA` row has a dangling `Parent` reference, making the output invalid for downstream tools that validate parent-child relationships (AGAT, gffread, IGV). (bug: GFF3 spec violation -- missing gene feature records)

https://github.com/GoekeLab/bambu/blob/67f184d736f5b8dca55b5f77b2b5c94a1c226580/R/readWrite.R#L320-L355

3. `writeAnnotationsToGFF3()` message says `"gtfs"` instead of `"gff3s"` and contains a typo `"th "` (should be `"the"`). This was copy-pasted from `writeAnnotationsToGTF()` without updating. (bug: misleading user-facing message referencing wrong format)

https://github.com/GoekeLab/bambu/blob/67f184d736f5b8dca55b5f77b2b5c94a1c226580/R/readWrite.R#L376-L378

cying111:
gff3 entries are not sorted, would suggest to follow sth similar to gtf entries? i.e., transcript entry, then followed by all exon entries sorted by chromosome coordinate, if need to add gene entry, gene entry can be on top of transcript entry

but I noted that writeToGTF output gtf entries are also not sorted, maybe need to change this as well?

One more thing when reading gff3 documentation, ID always comes before Parent, maybe also good to keep the same with the convention?

Missing IDs for exons, maybe can have sth as ID=exon_1, for all exons, so that rank is not necessary?

Screenshot 2026-04-17 at 11 10 04 AM

R/readWrite.R
```diff
@@ -17,23 +19,36 @@
...
-        if (!dir.exists(outdir))
-            dir.create(outdir, recursive = TRUE)
-
-        transcript_grList <- rowRanges(se)
-        prefix <- ifelse(prefix != "", paste0(prefix, "_"), "")
-        transcript_gtffn <- paste(outdir, prefix, sep = "")
-        gtf <- writeAnnotationsToGTF(annotation = transcript_grList,
-            file = transcript_gtffn, outputExtendedAnno = outputExtendedAnno, 
-            outputAll = outputAll, outputBambuModels = outputBambuModels, outputNovelOnly = outputNovelOnly)
+  } else if (!outputAnnFormat %in% c("gtf", "gff3")) {
```
cying111, line 28: maybe should also allow for gff? and assume gff meant for gff3?

R/readWrite.R
```diff
@@ -17,23 +19,36 @@
...
+    transcript_annfn <- paste(outdir, prefix, sep = "")
+    
+    if (outputAnnFormat == "gtf") {
+      gtf <- writeAnnotationsToGTF(annotation = transcript_grList,
+                                   file = transcript_annfn, outputExtendedAnno = outputExtendedAnno, 
+                                   outputAll = outputAll, outputBambuModels = outputBambuModels, outputNovelOnly = outputNovelOnly)
+    }
+    
+    if (outputAnnFormat == "gff3") {
+      gff3 <- writeAnnotationsToGFF3(annotation = transcript_grList,
```
cying111, line 46: writeAnnotationsToGFF3 and writeAnnotationsToGTF already write to files, so not need to assign to gff3, same for above, no need to assign to gtf, and both are not used in following lines as well 
```suggestion
      writeAnnotationsToGFF3(annotation = transcript_grList,
```

R/readWrite.R
```diff
@@ -247,6 +262,135 @@ writeAnnotationsToGTF <- function(annotation, file, geneIDs = NULL, outputExtend
...
+                 attributes = sub(";$", "", attributes))
+  
+  # Export
+  writeLines("##gff-version 3", file)
+  utils::write.table(gff3, file = file, append = TRUE, quote = FALSE, row.names = FALSE,
+                     col.names = FALSE, sep = "\t")
+}
+
+
+writeAnnotationsToGFF3 <- function(annotation, file, geneIDs = NULL, outputExtendedAnno = TRUE,
```
cying111, line 369: this function  maybe can be combined with writeAnnotationsToGTF to one writeAnnotations function, with one more parameter to call writeToGTF or writeToGFF3 later?

R/readWrite.R
```diff
@@ -247,6 +262,135 @@ writeAnnotationsToGTF <- function(annotation, file, geneIDs = NULL, outputExtend
...
+      geneIDs <- as_tibble(mcols(annotation, use.names = FALSE)[,
+                                                                c("TXNAME", "GENEID")])
+      geneIDs$seqnames <- unlist(unique(seqnames(annotation)))
+      geneIDs$strand <- unlist(unique(strand(annotation)))
+    }
+  }
+  
+  df <- left_join(df, geneIDs[, c("TXNAME", "GENEID")], 
+                  by = c("group_name" = "TXNAME"))
+
```
cying111, line 316: processing until this step is duplicated from writeToGTF and writeToGFF3, maybe can consider to combine 2 functions?

R/readWrite.R
```diff
@@ -247,6 +262,135 @@ writeAnnotationsToGTF <- function(annotation, file, geneIDs = NULL, outputExtend
...
+    mutate(attributes = sub("ID=", "Parent=", attributes))
+  
+  # Dataframe for transcripts
+  dfTx <- as_tibble(as.data.frame(range(ranges(annotation))))
+  dfTx <- left_join(dfTx, geneIDs, by = c("group_name" = "TXNAME"))
+  dfTx$group_name <- paste("ID=", dfTx$group_name, ";", sep = "")
+  dfTx$GENEID <- paste("Parent=", dfTx$GENEID, ";", sep = "")
+  
+  
+  if(!is.null(mcols(annotation)$NDR)) {
```
cying111, line 337: this if section is duplicated (appears 4 times)? maybe can modularise it as well?

R/readWrite.R
```diff
@@ -247,6 +262,135 @@ writeAnnotationsToGTF <- function(annotation, file, geneIDs = NULL, outputExtend
...
+  # Concat tx and exon, then drop the group_name
+  gff3 <- rbind(dfTx, dfExon) %>% group_by(group_name) %>%
+    arrange(as.character(seqnames), start) %>% ungroup() %>%
+    select(seqnames, source, feature, start, end, score,
+           strand, frame, attributes)
+  
+  # replace * with . and remove trailing ;
+  gff3 <- mutate(gff3, strand = recode_factor(strand, `*` = "."),
+                 attributes = sub(";$", "", attributes))
+
```
cying111, line 361: not sure where sorting should happen, I think maybe the step before mutate would be fine

cying111:
Code review from Claude code:

my feeling is that: issue 4 and 5 maybe also good to address, the usage of mRNA might also need to be changed, maybe can just be transcript? 


### 1. Copy-paste "gtfs" in GFF3 message — score 75

`R/readWrite.R:377`:

```r
message("The current NDR threshold already outputs all transcript models.
This may result in reduced precision for th extendedAnnotations and
supportedTranscriptModels gtfs")
```

Two problems: (a) "gtfs" should be "gff3s" — copy-paste leftover from
the GTF version at L249. (b) Pre-existing typo "th" → "the".

**Fix location:** `R/readWrite.R:377` — change "gtfs" to "gff3s" and
"th" to "the".

### 2. `Parent` before `ID` in transcript attributes — score 60

`R/readWrite.R:348`:

```r
attributes = paste0(GENEID, group_name, NDR, txScore, ...)
```

Where `GENEID = "Parent=<gene>;"` (L334) and `group_name = "ID=<tx>;"`
(L333). Produces `Parent=gene;ID=tx;...` — convention is `ID` first,
then `Parent`.

Verified against the [official GFF3 spec](https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md):
the spec does **not** mandate attribute order in column 9. It's
convention, not requirement. Most parsers (IGV, JBrowse) handle
either order. Downgraded from initial score of 85.

**Fix location:** `R/readWrite.R:348` — swap `GENEID` and `group_name`
in the `paste0()` call:

```r
attributes = paste0(group_name, GENEID, NDR, txScore, ...)
```

### 3. Missing gene-level records — score 60

`writeToGFF3()` writes only `mRNA` and `exon` records. Standard GFF3
practice (Ensembl, NCBI) uses a three-level hierarchy:
gene → mRNA → exon.

Verified against the [GFF3 spec](https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md)
and [NCBI GFF3 docs](https://www.ncbi.nlm.nih.gov/genbank/genomes_gff/):
gene features are "useful but NOT strictly required". The mRNA + exon
hierarchy is valid GFF3 on its own. The existing GTF writer
(`writeToGTF`) also skips gene records — consistent design choice.

Some genome browsers and validators may warn but will not reject the
file. No code change required unless broader tool compatibility is
needed.

### 4. Missing `@noRd` / `@export` on GFF3 functions — score 75

`writeToGFF3` (L266) and `writeAnnotationsToGFF3` (L369) have no
roxygen documentation at all. The parallel GTF functions (`writeToGTF`
L147, `writeAnnotationsToGTF` L241) are both `@export`ed with full
roxygen blocks.

Neither GFF3 function appears in NAMESPACE. If they are meant to be
internal-only, they should have `@noRd`. If users should be able to
call them directly (as with `writeToGTF()`), they need `@export` +
documentation.

**Fix location:** Add `#' @noRd` above `writeToGFF3` (L266) and
`writeAnnotationsToGFF3` (L369). Or, if public use is intended, add
full roxygen blocks matching the GTF equivalents and add to NAMESPACE.

### 5. Two independent `if` blocks instead of `if/else if` — score 50

`R/readWrite.R:56` and `R/readWrite.R:62`:

```r
if (outputAnnFormat == "gtf") {
    gtf <- writeAnnotationsToGTF(...)
}
if (outputAnnFormat == "gff3") {
    gff3 <- writeAnnotationsToGFF3(...)
}
```

Formats are mutually exclusive (validated at L45). Should be
`if/else if` to make exclusivity explicit and avoid a redundant
string comparison. Not a crash risk.

**Fix location:** `R/readWrite.R:62` — change `if` to `else if`.

## Pre-existing issues (faithfully reproduced from GTF writer)

These exist identically in the GTF writer and were copy-pasted into the
GFF3 version. Not introduced by this PR.

- **`novelTranscript` NA subsetting** (L388 / GTF L260):
  `annotation[mcols(annotation)$novelTranscript]` — if `novelTranscript`
  contains NAs, logical subsetting includes NA-indexed elements. Both
  versions have the same pattern; unlikely to trigger in practice since
  bambu always sets `novelTranscript` to TRUE/FALSE.

- **`recode_factor` deprecated** (L359 / GTF L214): dplyr 1.1.0+
  deprecates `recode_factor` in favor of `case_match()`. Will emit
  deprecation warnings on current dplyr.

- **`txScore` vs `maxTxScore` key naming asymmetry** (L297 exon-level
  `maxTxScore=` vs L339 transcript-level `txScore=`): These access
  different `mcols` fields (`$maxTxScore` vs `$txScore`), so the naming
  reflects the underlying data. Inconsistent but correct.

- **`mRNA` feature type for all transcripts** (L347): GFF3 uses `mRNA`
  for protein-coding transcripts. Non-coding RNA should use `transcript`
  or `ncRNA`. Using `mRNA` universally is technically incorrect per
  Sequence Ontology but common in practice. Design choice.

R/readWrite.R
hafiz-ismail: to do in the future as part of deduplicating code

R/readWrite.R
hafiz-ismail: to do in the future as part of deduplicating code

R/readWrite.R
hafiz-ismail: the column assignment is slightly different from the one at line 287

i.e. lines 296-302
```
    df$NDR <- paste("NDR=", as.character(NDR), ";", sep = "")
    df$txScore <- paste("maxTxScore=", as.character(txScore), ";", sep = "")
    df$txScore.noFit <- paste("maxTxScore.noFit=", as.character(txScore.noFit), ";", sep = "")
    df$novelGene <- paste("novelGene=", as.character(novelGene), ";", sep = "")
    df$novelTranscript <- paste("novelTranscript=", as.character(novelTranscript), ";", sep = "")
    df$txClassDescription <- paste("txClassDescription=", as.character(txClassDescription), ";", sep = "")
  }
```

vs lines 338-344
```
    dfTx$NDR <- paste("NDR=", as.character(mcols(annotation)$NDR), ";", sep = "")
    dfTx$txScore <- paste("txScore=", as.character(mcols(annotation)$txScore), ";", sep = "")
    dfTx$txScore.noFit <- paste("txScore.noFit=", as.character(mcols(annotation)$txScore.noFit), ";", sep = "")
    dfTx$novelGene <- paste("novelGene=", as.character(mcols(annotation)$novelGene), ";", sep = "")
    dfTx$novelTranscript <- paste("novelTranscript=", as.character(mcols(annotation)$novelTranscript), ";", sep = "")
    dfTx$txClassDescription <- paste("txClassDescription=", as.character(mcols(annotation)$txClassDescription), ";", sep = "")
  }
```
likewise the same block of code is also 'duplicated' in writeToGTF.
nevertheless, they are definitely very similar, can be rewritten and also modularized in the future

PR #586, opened 2026-08-13
Store combined transcript models in long form to bound peak memory

### LINKED ISSUE #587: Bound peak memory in combineSplicedTranscriptModels() by keeping combined models in long form
Tracking issue for the community pull request #586 by @dariansal.

## Problem

`combineSplicedTranscriptModels()` builds one wide table with a `start`, `end` and `readCount` column per sample. A read class appears in only a few samples, so the table is mostly `NA`, and the `full_join` fold holds a second copy while the first is still live.

Peak memory therefore scales with rows x samples instead of with the data. Large sample sets can exhaust memory before the step completes.

Example. Three intron chains across four samples. Only 6 of the 12 `(chain, sample)` cells were observed.

```
chain          st.s1 en.s1 rc.s1 | st.s2 en.s2 rc.s2 | st.s3 en.s3 rc.s3 | st.s4 en.s4 rc.s4
chr1:100-200     100   205     5 |    NA    NA    NA |   102   205     1 |   101   207     3
chr1:400-600      NA    NA    NA |   400   650     4 |    NA    NA    NA |    NA    NA    NA
chr2:50-90        50    95     2 |    52    95     6 |    NA    NA    NA |    NA    NA    NA
```

The table holds 36 cells, and 18 of them are `NA`. The fold that builds it:

```r
t <- s1
t <- full_join(t, s2)   # old t and new t are both live here
t <- full_join(t, s3)
t <- full_join(t, s4)
```

How the two forms scale:

```
                        3 chains x 4 samples      1e6 chains x 100 samples
wide cells              36                        3.0e8   (~2.4 GB, doubles)
wide cells that are NA  18   (50%)                2.9e8   (~97%)
long cells              18                        2.7e7   (~0.2 GB)
peak during the fold    2 x wide                  ~4.8 GB
```

## Proposed change

**Step 1. Keep only the populated `(chain, sample)` cells, in long form.**

```
chain          sample  start  end  readCount
chr1:100-200   s1        100  205          5
chr1:100-200   s3        102  205          1
chr1:100-200   s4        101  207          3
chr1:400-600   s2        400  650          4
chr2:50-90     s1         50   95          2
chr2:50-90     s2         52   95          6
```

This holds 18 cells and no `NA`.

**Step 2. Reduce by chain, in place of the join fold.**

```r
long %>% group_by(chain) %>% summarise(...)
```

No second copy of the table is live at any point.

**Step 3. Replace the expand-by-weight step in `readCountWeightedMedian()` with a cumulative-weight lookup.**

For integer weights, the type-7 median snapped up to the nearest observed value is the element at `floor(N/2)+1` of the expanded sorted vector, which is locatable from the cumulative weights alone.

Current behaviour, for the `start` of `chr1:100-200`:

```
values  = 100, 102, 101
weights =   5,   1,   3

expand:  100 100 100 100 100 101 101 101 102     <- 9 elements allocated
type-7 median of 9 = 100.0
snap up to the nearest observed value >= 100.0  ->  100
```

The replacement, same input:

```
sort by value:   100   101   102
weights:           5     3     1
cumulative:        5     8     9      N = 9

target = floor(9/2) + 1 = 5
first value whose cumulative >= 5  ->  100
```

Both give the same result on all three chains:

```
chain          values       weights   expand + median   cumulative lookup
chr1:100-200   100,102,101  5,1,3     100               100
chr1:400-600   400          4         400               400
chr2:50-90     50,52        2,6        52                52
```

The allocation this removes:

```
one chain with readCount 5000 in each of 100 samples
  expand -> a vector of 500,000 doubles, discarded immediately after
  lookup -> 100 cumulative sums
```

**Step 4. Retain the current implementation as `combineSplicedTranscriptModelsDense()`.**

Use it whenever the sparse path cannot guarantee an identical result:

- fewer than three samples (`nGroups == 1`, no join happens, the `NSample*` columns stay logical)
- any input repeating an intron chain within one sample, which the join would fan out

**Step 5. Keep the RNG draw where it is,** so both paths see the same grouping for a given seed.

dariansal:
## Problem

`combineSplicedTranscriptModels()` builds one table carrying a `start`, `end` and `readCount` column per sample. A read class is observed in only a few samples, so that table is mostly `NA`, and the `full_join` fold materialises a second copy while the first is still live.

Peak memory therefore grows with rows × samples rather than with the data, and large sample sets can exhaust memory before the step completes.

## Change

Keep only the populated `(chain, sample)` cells in long form and reduce them by chain.

The reduction also avoids expanding values by their weights. `readCountWeightedMedian()` repeats each value `readCount` times, takes the type-7 median and snaps up to the nearest observed value; with integer weights that is exactly the element at `floor(N/2)+1` of the expanded sorted vector, locatable from the cumulative weights alone.

## Output is unchanged

The existing implementation is kept as `combineSplicedTranscriptModelsDense()` and runs whenever the sparse path cannot guarantee an identical result:

- fewer than three samples, where `nGroups` is 1, no join happens, and the `NSample*` columns stay logical
- any input repeating an intron chain within one sample, which the join would fan out

The RNG draw stays where it was, so both paths see the same grouping for a given seed.

## Verification

- `upperMedianByGroup()` against `readCountWeightedMedian()` across 50 randomised configurations
- both paths `identical()` on real read classes at n=20 and n=40, RNG seeded the same for each; the combine step ran 42.6× and 44.4× faster
- both paths against each other on synthetic read classes, plus the duplicate-chain fallback

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,8 @@
+bambu v3.13.2 (change date: 2026-08-13)
+==========================================
+Minor fixes:
+    * Reduced peak memory in combineSplicedTranscriptModels by storing the per-sample start, end and readCount values in long form rather than as one column trio per sample. Output is unchanged; the previous implementation is retained and used whenever the sparse path cannot guarantee an identical result.
+
 bambu v3.2.0 (change date: 2023-Apr-26)
 ==========================================
 Minor fixes:
--- a/R/bambu-extendAnnotations-utilityCombine.R
+++ b/R/bambu-extendAnnotations-utilityCombine.R
@@ -31,38 +31,237 @@ isore.combineTranscriptCandidates <- function(readClassList,
 
 
 #' combine spliced transcript models
+#'
+#' Dispatches to a sparse reduction, falling back to the original dense
+#' implementation when the sparse path cannot reproduce it exactly (see
+#' combineSplicedTranscriptModelsSparse). Both return the same 12 columns,
+#' in the same order, with the same row order.
 #' @noRd
-combineSplicedTranscriptModels <- function(readClassList, bpParameters, 
-        min.readCount, min.readFractionByGene, min.txScore.multiExon, 
+combineSplicedTranscriptModels <- function(readClassList, bpParameters,
+        min.readCount, min.readFractionByGene, min.txScore.multiExon,
         min.txScore.singleExon, verbose){
     bpParameters$progressbar = FALSE
     options(scipen = 999) #maintain numeric basepair locations not sci.notfi.
     start.ptm <- proc.time()
     n_sample <- length(readClassList)
-    nGroups = max(ceiling(n_sample/10),min(bpworkers(bpParameters), 
+    nGroups = max(ceiling(n_sample/10),min(bpworkers(bpParameters),
                                             round(n_sample/2)))
     indexList <- sample(rep(seq_len(nGroups), length.out=n_sample))
     indexList <- splitAsList(seq_len(n_sample), indexList)
-    combinedFeatureTibbleList <- bplapply(seq_along(indexList), function(g){
-        indexVec <- indexList[[g]]
-        return(sequentialCombineFeatureTibble(readClassList[indexVec],
-            indexVec, intraGroup = TRUE, 
-            min.readCount = min.readCount, 
-            min.readFractionByGene = min.readFractionByGene, 
-            min.txScore.multiExon = min.txScore.multiExon,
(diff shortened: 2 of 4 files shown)


lingminhao:
Hi @dariansal , thanks for creating the PR. It may take some time for internal review and we usually gather all community feedback and addresses all of them during a major / minor release of next version of Bambu. We will let you know here if the PR is accepted and passes the internal review process.

For the moment, feel free to use this branch for your own usage. Thanks again for contributing!

Issue #581, opened 2026-06-18
Difficulties running on AWS HealthOmics


SBrennan-MBS:
Hi guys,

Thanks very much for making this tool available for all of us! I am trying to run bambu as a private workflow on AWS HealthOmics, but I am hitting a consistent error:

Warning: Bioconductor online version validation disabled; see	?BIOCONDUCTOR_ONLINE_VERSION_DIAGNOSIS	[2026-06-18 01:59:02] ERROR in bambu(): error in evaluating the argument 'table' in selecting a method for function '%in%': Bioconductor version map cannot be validated; is it misconfigured?

I have disabled BIOCONDUCTOR_ONLINE_VERSION_DIAGNOSIS and even tried patching in the bioconductor config.yaml in both my R script and my nextflow.config file. From what I can tell, nextflow is passing the reference files and bam files as expected, but because I have >10 samples, a number of child R processes are spinning up and they drop both the online diagnosis and config.yaml. Then the child processes attempt online validation, can't connect to the internet, and fail.

Is there any way around this?

Cheers

lingminhao:
Hi @SBrennan-MBS,

Thanks for considering bambu as part of your workflow on AWS HealthOmics. To confirm a few things, are you currently providing the genome as a file path, and is the HealthOmics environment effectively offline (no outbound internet access)? If so, that might explain the failure. bambu runs a quick check, `BSgenome::available.genomes()` when genome is provided as a path that needs online access.

One way around this is to pass the genome as an `FaFile` object instead of a path. When the genome isn't a plain string, bambu skips the `available.genomes()` lookup entirely, so no network call is needed (just make sure the `.fai` index also sits next to the FASTA):

```r
library(Rsamtools)
indexFa("/path/to/genome.fa")          
genome <- FaFile("/path/to/genome.fa") # a temporary fix here
bambu(reads = bams, annotations = annotations, genome = genome)
```

If you can confirm this gets your run through on HealthOmics, we'll work on a proper fix for a future release so that providing a FASTA path skips the online lookup automatically.

Hope it helps!
Min Hao

SBrennan-MBS:
Hi @lingminhao,

Thanks very much for your quick reply on this. Yeah HealthOmics only has access to AWS S3 by default so no general internet access. I have been looking into other options myself and I could set up a VPC connection so that my HealthOmics instances can have internet access, but would be very pleased if there is an easier option.

[...]

PR #588, opened 2026-09-02
Fix bambu requiring internet access when given a local FASTA genome path

### LINKED ISSUE #339: Trying to use bambu on an offline cluster
Hello, I am trying to run the new version of bambu (3.0.3, R 4.2.2) on our cluster and am getting this error. BiocManager documentation is telling me about the possibility of having an offline yaml but I am not sure how to use it in this case (https://cran.r-project.org/web/packages/BiocManager/vignettes/BiocManager.html#offline-config.yaml) 
```
--- Start generating read class files ---
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'table' in selecting a method for function '%in%': Bioconductor version cannot be validated; no internet connection?  See
  #troubleshooting section in vignette
Calls: bambu ... <Anonymous> -> .version_validate -> ifelse -> .stop
In addition: Warning messages:
1: In file(con, "r") :
  URL 'https://bioconductor.org/config.yaml': status was 'Couldn't connect to server'
2: In file(con, "r") :
  URL 'http://bioconductor.org/config.yaml': status was 'Couldn't connect to server'
Execution halted
```
Thank you for the tool!
Sofia 

edit: previous version of bambu with R 4.1.2 worked fine.

jonathangoeke:
## Summary
- Fixes #339: `bambu()` failed on offline machines (e.g. an offline HPC cluster, or a self-contained biocontainers image with no internet egress — see [comment](https://github.com/GoekeLab/bambu/issues/339#issuecomment-5470632146)) even when given a plain local FASTA file path for `genome`.
- Root cause: `checkInputSequence()` in `R/bambu_utilityFunctions.R` unconditionally called `BSgenome::available.genomes()` for any character `genome` argument, before ever checking whether the string was simply a local file path. That call reaches out to Bioconductor's package repository over the network, so it threw a raw Bioconductor/BiocManager error instead of just reading the FASTA.
- Fix: only attempt the `BSgenome::available.genomes()` / `getBSgenome()` lookup when the string is **not** an existing local file, and wrap that lookup in `tryCatch` so a blocked/failed network call falls through to the existing FASTA-read path (and its existing clear error message) instead of crashing.

## Test plan
- [x] `devtools::load_all()` + traced `BSgenome::available.genomes()` to confirm it is no longer called when `checkInputSequence()` is given the package's local test FASTA (`inst/extdata/Homo_sapiens.GRCh38.dna_sm.primary_assembly_chr9_1_1000000.fa`)
- [x] Confirmed a nonexistent path still produces the original `"Input genome file not readable."` error
- [x] `testthat::test_file()` passing locally for `test_bambu_arguments.R`, `test_scoreRC.R`, `test_isore.R` (all exercise `checkInputSequence()`/`bambu()` with a local FASTA path)
- [x] `test_setNDR.R` fails identically before and after this change (missing fixture path in this environment) — confirmed unrelated to this fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/R/bambu_utilityFunctions.R
+++ b/R/bambu_utilityFunctions.R
@@ -147,11 +147,16 @@ checkInputs <- function(annotations, reads, readClass.outputDir, genomeSequence)
 checkInputSequence <- function(genomeSequence) {
     if (is.null(genomeSequence)) stop("Reference genome sequence is missing,
         please provide fasta file or BSgenome name, see available.genomes()")
+    if (is.character(genomeSequence) && !file.exists(genomeSequence)) {
+        isBSgenomeName <- tryCatch(
+            genomeSequence %in% BSgenome::available.genomes(),
+            error = function(cond) FALSE)
+        if (isBSgenomeName) {
+            genomeSequence <- BSgenome::getBSgenome(genomeSequence)
+            return(genomeSequence)
+        }
+    }
     if(is.character(genomeSequence)){
-    if (genomeSequence %in% BSgenome::available.genomes()) {
-        genomeSequence <- BSgenome::getBSgenome(genomeSequence)
-        return(genomeSequence)
-    } 
     tryCatch(
     {
         if (.Platform$OS.type == "windows") {


cying111 (approved):
the change looks good to me and I have tested it offline where local fasta files are loaded, before the change there is an error similar like the one described in the issue, after the change it can run smoothly

lingminhao:
same as community issue #581, can inform the user once it's approved.

